PR #85, opened 2017-02-15
Allow multiple version in constructor of ApiVersionAttribute

pkiers:
I extended the ApiVersionAttribute to support multiple versions, this was already supported in the base class ApiVersionsBaseAttribute.

### DIFF
--- a/src/Common/ApiVersionAttribute.cs
+++ b/src/Common/ApiVersionAttribute.cs
@@ -20,25 +20,6 @@ namespace Microsoft.AspNetCore.Mvc
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
         /// </summary>
-        /// <param name="version">The <see cref="ApiVersion">API version</see>.</param>
-        protected ApiVersionAttribute( ApiVersion version )
-            : base( version )
-        {
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
-        /// </summary>
-        /// <param name="versions">An <see cref="Array">array</see> of API version strings.</param>
-        [CLSCompliant( false )]
-        public ApiVersionsBaseAttribute( params string[] versions )
-            : base ( versions )
-        {
-        }
-
-        /// <summary>
-        /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
-        /// </summary>
         /// <param name="version">The API version string.</param>
         public ApiVersionAttribute( string version )
             : base( version )
@@ -48,8 +29,26 @@ namespace Microsoft.AspNetCore.Mvc
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
         /// </summary>
+        /// <param name="versions">An <see cref="Array">array</see> of API version strings.</param>
+        [CLSCompliant( false )]
+        public ApiVersionAttribute( params string[] versions )
+            : base ( versions )
+        {
+        }
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
+        /// </summary>
+        /// <param name="version">The <see cref="ApiVersion">API version</see>.</param>
+        protected ApiVersionAttribute( ApiVersion version )
+            : base( version )
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
+        /// </summary>
         /// <param name="versions">An <see cref="Array">array</see> of <see cref="ApiVersion">API versions</see>.</param>
-        protected ApiVersionsBaseAttribute( params ApiVersion[] versions )
+        protected ApiVersionAttribute( params ApiVersion[] versions )
             : base ( versions )
         {
         }
--- a/src/Common/ApiVersionAttribute.cs
+++ b/src/Common/ApiVersionAttribute.cs
@@ -29,9 +29,28 @@ namespace Microsoft.AspNetCore.Mvc
         /// <summary>
         /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
         /// </summary>
+        /// <param name="versions">An <see cref="Array">array</see> of API version strings.</param>
+        [CLSCompliant( false )]
+        public ApiVersionsBaseAttribute( params string[] versions )
+            : base ( versions )
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
+        /// </summary>
         /// <param name="version">The API version string.</param>
         public ApiVersionAttribute( string version )
             : base( version )
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
+        /// </summary>
+        /// <param name="versions">An <see cref="Array">array</see> of <see cref="ApiVersion">API versions</see>.</param>
+        protected ApiVersionsBaseAttribute( params ApiVersion[] versions )
+            : base ( versions )
         {
         }
 


JayBazuzi:
I vote against this, but I do want to understand why you want it.

Tell us more about your situation.

pkiers:
Our current setup is that we run version X of our API when in a release we have a breaking version the whole API goes to next version. So all the none breaking endpoints get a new ApiVersion. So if we have version 12 active we get on the non-changed controllers this:

```
[ApiVersion("1.0")]
[ApiVersion("2.0")]
[ApiVersion("3.0")]
[ApiVersion("4.0")]
[ApiVersion("5.0")]
[ApiVersion("6.0")]
[ApiVersion("7.0")]
[ApiVersion("8.0")]
[ApiVersion("9.0")]
[ApiVersion("10.0")]
public class MyController {}
```
versus
```
[ApiVersion("1.0", "2.0", "3.0", "4.0", "5.0", "6.0", "7.0", "8.0", "9.0", "10.0")]
public class MyController {}
```
To me that feels much cleaner, but the what I actually wanted to build in our project was this:
```
public class ApiVersionRangeAttribute : ApiVersionAttribute
{
    protected ApiVersionAttribute( string fromVersion, string toVersion )
        : base( SomeStaticVersionClass.GetVersionArrayFrom(fromVersion, toVersion) )
    {
    }
}

[ApiVersionRange("2.0", "10.0")]
public class MyController {}


[ApiVersionRange("2.0")] // Which would mean from 2.0 to latest
public class MyController {}
```
That way we only can copy the old controller or action to a V?? folder and add a new version and don't have to go thru all the controllers / actions and add new attributes.

The SomeStaticVersionClass could be a generated .cs file that just contains array with all versions that have been published to customers.

commonsensesoftware:
There are two points of discussion here.  Let me first start by talking about defining multiple API versions. In previous, internal versions of this library, it was possible to define the API versions as you are proposing. Adoption of the libraries exposed confusion with respect to the intention of the attributes; names matter. Several refactorings occurred to get us to where things are now. The revised design went something like this:

* ApiVersionAttribute - "has API version" and should be singular
* ApiVersionsAttribute - was considered, but deemed unnecessary; it's also confusing with only 1 version
* Most services have versioning policies and the most common versioning policy is to support _N-2_ versions, which means the version list will be relatively small, even if it's something like _N-5_

I also recognize that it's possible some services may have a very large list of versions out in the wild. To that end, we wanted to make it really easy to extend or otherwise customize the library to suite service authors' particular needs without requiring a ton of work. This is the primary reason for the **ApiVersionsBaseAttribute**.

Ultimately, the API version discovery and aggregation methods do not rely on attributes. The discovery mechanisms look for and use implementations of [IApiVersionProvider](../blob/master/src/Common/Versioning/IApiVersionProvider.cs).  Attributes just happen to be one of the approaches that provide API versions.  You can easily use the provided base attribute or roll your own and implement **IApiVersionProvider**. An example of this capability is illustrated in the [extensions and customizations](../wiki/Extensions-and-Customizations) topic.

commonsensesoftware:
Supporting API version ranges has been asked for in the past.  This isn't something that can really be supported due the expected runtime behaviors.  The controller and action selection process must be unambiguous.  In addition, if ranges are supported, it's becomes difficult, impossible, or at the very least, very complicated to support sun-setting API versions (e.g. the complete removal).

Imagine you had something like: `[ApiVersionRange( "1.0", "3.0" )]`, but API version **2.9-Beta** is now completely removed (remember that deprecated is still reachable).  This API version would fall into the range. I'm just touching the tip of the iceberg here, but hopefully you're seeing how it can get very, very complex quickly.

In practice, what service authors really want is a way to:

* Centrally manage their API versions; especially when the versions are consistent for an entire codebase
* Avoid changing or otherwise touching code unnecessarily
* Avoid entering large lists of values (e.g. API versions)

These are all admirable and we want to enable these scenarios, but ultimately they are _preferences_. There are several alternatives that can make this work for you:

* Use the [conventions](../wiki/API-Version-Conventions) support instead of attributes. You can define your API versions statically or via configuration and then apply them at start up. For example, `configuration.AddApiVersioning( o => o.Conventions.Controller<MyController>().HasApiVersions( AllSupportedApiVersions )`. You also use an extension method to simplify the setup; something like _ImplementsStandardApiVersions_.
* Use a custom attribute that inherits from **ApiVersionsBaseAttribute** that includes all of your common/default API versions. This would allow you centrally manage most API versions while still having the flexibility to add additional API versions at specific controller levels.
* Use a custom attribute that implements **IApiVersionProvider** and resolve the configured API versions using some type of _Service Locator_. This could be simply done for all API versions (`[ConfiguredApiVersions]`) or for controller specific API versions (`[ConfiguredApiVersions( typeof( MyController ) )]`).
* Since you mentioned a **V??** folder, that should mean that there is a corresponding namespace to the folder. If you define a convention for the namespace name, then you can use a custom attribute that implements **IApiVersionProvider** and builds up the API version from that. For example, `[ApiVersionFromNamespace( typeof( MyController ) )]` where the full type name is "Services.V2_5.MyController". The "V2_5" can be extracted from the namespace and parsed into `new ApiVersion( 2, 5 )`.
* Use a custom attribute that inherits from **ApiVersionsBaseAttribute** that can build up an explicit range. For example, `[ApiVersionRange( From = 1, To = 3, Step = 1 )]` where _Step_ likely defaults to 1 and is optional. This would generate and supply the API versions 1.0, 2.0, and 3.0 to the base implementation.

As you can see, there are a **lot** of choices and these can all be done without any changes to the existing implementation of the libraries.  Choosing the right fit for you services is up to you.

pkiers:
@commonsensesoftware Well the first thing I did was create own attribute that extended ApiVersionsBaseAttribute.
```
[AttributeUsage(AttributeTargets.Class, AllowMultiple = true, Inherited = false)]
public partial class ApiVersionsAttribute : ApiVersionsBaseAttribute, IApiVersionProvider
{
    /// <summary>
    /// Initializes a new instance of the <see cref="ApiVersionAttribute"/> class.
    /// </summary>
    /// <param name="version">The API version string.</param>
    public ApiVersionsAttribute(params string[] versions)
        : base(versions)
    {
    }

    bool IApiVersionProvider.AdvertiseOnly => false;

    /// <summary>
    /// Gets or sets a value indicating whether the specified set of API versions are deprecated.
    /// </summary>
    /// <value>True if the specified set of API versions are deprecated; otherwise, false.
    /// The default value is <c>false</c>.</value>
    public bool Deprecated { get; set; }
}
```
Then I used this
```
[ApiVersions("1.0", "2.0")]
[Route("v{version:apiVersion}/templates")]
[EnableCors("AllowAllOrigins")]
public class TemplatesController : BaseController { }
```
But then my endpoint returns:
```
{
  "Error": {
    "Code": "UnsupportedApiVersion",
    "Message": "The HTTP resource that matches the request URI 'http://api.sitebuilder.dev:5000/v2.0/templates' does not support the API version '2.0'."
  }
}
```
While the following does work
```
[ApiVersion("1.0")]
[ApiVersion("2.0")]
[Route("v{version:apiVersion}/templates")]
[EnableCors("AllowAllOrigins")]
public class TemplatesController : BaseController { }
```
So even when I am implementing IApiVersionProvider and extending ApiVersionsBaseAttribute the routes don't work.

But if you say that this should work I will do some digging and see where in the code it goes wrong.

JayBazuzi:
BTW, if you decide to use a list of versions, this is slightly more compact:

    [ApiVersion("1.0"), ApiVersion("2.0")]

pkiers:
Yeah, well the full idea is have a apisettings.json configuration that looks like this.
```
"ApiVersions": {
  "Versions": [
    "1.0",
    "1.1",
    "2.0",
    "3.0-alpha"
  ],
  "LatestApiVersion": "2.0",
  "PreviewApiVersion": "3.0-alpha"
}
```
And try implement a range attribute with [ApiVersions(From = "1.1", IncludePreview = true)] or [ApiVersions(From = "1.0", To = "2.0")]. Then using the Versions array to specify what range of versions the API contains. So if I can figure out why the ApiVersionsAttribute I posted above is not working and fixed that I am all fine 👍

commonsensesoftware:
Your attribute implementation looks correct. You even nailed implementing the illusive **IApiVersionProvider**, which **ApiVersionsBaseAttribute** does not actually implement (intentionally).

I took an example I was using to help someone yesterday, added the custom **ApiVersionsAttribute**, using the same implementation you have, and it works. I've attached the solution for your convenience to compare against ([WebApiVersionExample.zip](https://github.com/Microsoft/aspnet-api-versioning/files/780831/WebApiVersionExample.zip)).


It would seem there is something wrong with your routing configuration, though it's a little odd if it works without the custom attribute; otherwise, there must be some other side effecting configuration like CORS. I'm not sure why CORS would matter since that middleware runs so early in the pipeline.  If you can share your start-up configuration and/or the solution (assuming it's small and sharable), I don't mine giving it a once-over.

pkiers:
The project is rather big and am not allowed to share it either, but I took your example and added a clean .net core mvc project and copied over attribute and changed the controller some to make everything work. But then also the custom attribute is not working.
```
[ApiVersions( "2.0", "4.0" )]
[Route( "api/v{version:apiVersion}/tfmemployee" )]
public class TfmEmployee2Controller : Controller
{
    [Route("")]
    public ObjectResult Get() => Ok($"Success (from {GetType().Name})");
}
```
Project: [WebApiVersionExample.zip](https://github.com/pkiers/aspnet-api-versioning/raw/master/WebApiVersionExample.zip)

commonsensesoftware:
OK ... I didn't realize this was ASP.NET Core. A lot of effort was made to try and keep things as similar as possible between the platforms, but there are inevitable differences.  In this particular case, the issue is that in addition to implementing the **IApiVersionProvider**, you must also implement **IControllerModelConvention**. The new convention model in ASP.NET Core allows providing metadata without relying solely on attributes. The API Versioning libraries used a similar design for discovering the version metadata.

I wish I could say that was the end of this thread and all was good in the world again. Unfortunately, that's the not the case. I regret to report that you are the _proud_ winner of a design bug. Congratulations! It turns out that you can not implement **IControllerModelConvention** in the same manner as **ApiVersionAttribute** because the constructors on **ApiVersionModel** required to do so are `internal`.

I always worried this could cause an issue. The reason most of the constructors for the **ApiVersionModel** are internal is because of the way things get aggregated and used under the covers. It's not exactly clear to the outside world as to which constructor to use when. While there are some (clear) exposed uses and extension methods, most were kept hidden.  Clearly, that was a mistake because it has broken the extensibility model. I've filed a new bug (Issue #87). My recommendation is to open up just the constructors that are need for extensibility. If either of you have additional thoughts or suggestions on the matter, let's track them there so the rest of the community can choose to chime in.

The fix isn't terribly difficult, but it may take a day or two to get out. In the mean time, here is a **horrible**, **horrible** hack that will work:

```c#
using Microsoft.AspNetCore.Mvc.ApplicationModels;
using Microsoft.AspNetCore.Mvc.Versioning;
using System;
using System.Linq;
using System.Reflection;
using static System.Linq.Expressions.Expression;
using static System.Reflection.BindingFlags;

[AttributeUsage( AttributeTargets.Class, AllowMultiple = true, Inherited = false )]
public sealed class ApiVersionsAttribute : ApiVersionsBaseAttribute, IApiVersionProvider, IControllerModelConvention
{
    static readonly Func<ControllerModel, ApiVersionModel> NewApiVersionModelForController;
    static readonly Func<ControllerModel, ActionModel, ApiVersionModel> NewApiVersionModelForAction;

    static ApiVersionsAttribute()
    {
        var type = typeof( ApiVersionModel );
        var ctors = type.GetConstructors( Instance | NonPublic );
        var ctor = ctors.Single( c => c.GetParameters()
                                       .Select( p => p.ParameterType )
                                       .SequenceEqual( new[] { typeof( ControllerModel ) } ) );
        var controller = Parameter( typeof( ControllerModel ), "controller" );
        var action = Parameter( typeof( ActionModel ), "action" );
        var body = New( ctor, controller );

        NewApiVersionModelForController = Lambda<Func<ControllerModel, ApiVersionModel>>( body, controller ).Compile();

        ctor = ctors.Single( c => c.GetParameters()
                                   .Select( p => p.ParameterType )
                                   .SequenceEqual( new[] { typeof( ControllerModel ), typeof( ActionModel ) } ) );
        body = New( ctor, controller, action );

        NewApiVersionModelForAction = Lambda<Func<ControllerModel, ActionModel, ApiVersionModel>>( body, controller, action ).Compile();
    }

    public ApiVersionsAttribute( params string[] versions ) : base( versions ) { }

    bool IApiVersionProvider.AdvertiseOnly => false;

    public bool Deprecated { get; set; }

    public void Apply( ControllerModel controller )
    {
        //controller.SetProperty( new ApiVersionModel( controller ) );

        //foreach ( var action in controller.Actions )
        //{
        //    action.SetProperty( controller );
        //    action.SetProperty( new ApiVersionModel( controller, action ) );
        //}

        // HACK: design flaw workaround until fixed version is published
        controller.SetProperty( NewApiVersionModelForController( controller ) );

        foreach ( var action in controller.Actions )
        {
            action.SetProperty( controller );
            action.SetProperty( NewApiVersionModelForAction( controller, action ) );
        }
    }
}
```

pkiers:
Just for my understanding why is the ApiVersionAttribute class working without implementing IControllerModelConvention? My custom attribute is exactly the same they just have different constructors.

commonsensesoftware:
The **ApiVersionAttribute** class does implement **IControllerModelConvention**. Since there are variances between the platforms, the source is split with partial classes.  Make sure you're looking at both source files in [ApiVersionAttribute.cs (Common)](../blob/master/src/Common/ApiVersionAttribute.cs) and [ApiVersionAttribute.cs (ASP.NET Core)](../blob/master/src/Microsoft.AspNetCore.Mvc.Versioning/ApiVersionAttribute.cs).

pkiers:
Ahh, that I did not catch! Anyway thanks for the support, we got it up and running how I wanted it and will replace the hack once the issue has been released. You can close the PR it you want :)

commonsensesoftware:
Closing this PR. The root issue will be resolved by #87.

PR #96, opened 2017-03-06
Refactor Request Properties

commonsensesoftware:
Coalesces the various API versioning request properties into a single object. This improves overall lookup performance and reduces the number of entries in the property bag of the current request. Superfulous extension methods that are supplanted by this refactoring have also been removed.


JayBazuzi:
I want my definition of refactoring back.

commonsensesoftware:
_[hands you the definition back]_

You could argue that [refactor](https://en.wikipedia.org/wiki/Code_refactoring) hasn't be followed in the strictest form; however, there are already required, breaking changes for the next release. These likely only affect internal use and extenders. In most cases, there will be no visible change. In each case, the capability or feature has been changed as opposed to removed. Each of these will be called out in the release notes.

I'd like to condense as many of these to a single release as possible so there isn't a moving target for adopters.  As always, I value your opinion.

PR #100, opened 2017-03-14
Update API Versioning for OData to Support 6.0

commonsensesoftware:
Updates the API Versioning implementation for OData to incorporate the changes from the Web API OData 6.0 library.


triplebeta:
Can it be that the package specification was not yet updated?
Package https://www.nuget.org/packages/Microsoft.AspNet.OData.Versioning/2.1.0-beta2

specifies the following dependencies:
Microsoft.AspNet.WebApi.Versioning (>= 2.1.0-beta1)
Microsoft.AspNet.OData (>= 5.9.1 && **< 6.0.0**)

According to your fix I expected to have become <= 6.0.0

commonsensesoftware:
The 6.0 version has not been released in a package - yet. You can see that  [Microsoft.AspNet.OData.Versioning.csproj (Line 20)](../blob/master/src/Microsoft.AspNet.OData.Versioning/Microsoft.AspNet.OData.Versioning.csproj#L20) does show the 6.0.0 reference as expected; however, I have not published a new package with this included.

There are breaking changes between 5.9.x and 6.0.0. I'm not planning on maintaining both variants. At the same time, there has been an equally long ask for Swagger support which is almost ready as well. I was trying to roll everything out at same, or almost the same, time so there is a clear path forward. It would hopefully provide a way back if needed. The OData Swagger support is taking longer than expected. There have been a lot of recent asks this, so I think I may just make publish another round of beta releases that will include the OData 6.0 support and Swagger support for ASP.NET Web API and Core.

triplebeta:
Thanks Chris, good to know!

 

Van: Chris Martinez [mailto:notifications@github.com] 
Verzonden: Sunday, April 9, 2017 18:15
Aan: Microsoft/aspnet-api-versioning <aspnet-api-versioning@noreply.github.com>
CC: J.Bijleveld <Jeroen@TripleBeta.nl>; Comment <comment@noreply.github.com>
Onderwerp: Re: [Microsoft/aspnet-api-versioning] Update API Versioning for OData to Support 6.0 (#100)

 

The 6.0 version has not been released in a package - yet. You can see that Microsoft.AspNet.OData.Versioning.csproj (Line 20) does show the 6.0.0 reference as expected; however, I have not published a new package with this included.

There are breaking changes between 5.9.x and 6.0.0. I'm not planning on maintaining both variants. At the same time, there has been an equally long ask for Swagger support which is almost ready as well. I was trying to roll everything out at same, or almost the same, time so there is a clear path forward. It would hopefully provide a way back if needed. The OData Swagger support is taking longer than expected. There have been a lot of recent asks this, so I think I may just make publish another round of beta releases that will include the OData 6.0 support and Swagger support for ASP.NET Web API and Core.

PR #103, opened 2017-03-25
Add Support for IApiExplorer

commonsensesoftware:
This contains the first reviewable set of changes to enable IApiExplorer for API-versioned services in ASP.NET Web API.  The next iteration will contain support for ASP.NET Core.


samples/aspnetcore/SwaggerSample/ImplicitApiVersionParameter.cs
```diff
@@ -0,0 +1,72 @@
...
+    using Microsoft.AspNetCore.Mvc.ApiExplorer;
+    using Swashbuckle.AspNetCore.Swagger;
+    using Swashbuckle.AspNetCore.SwaggerGen;
+    using System.Collections.Generic;
+    using System.Linq;
+
+    /// <summary>
+    /// Represents the Swagger/Swashbuckle operation filter used to document the implicit API version parameter.
+    /// </summary>
+    public class ImplicitApiVersionParameter : IOperationFilter
```
xperiandri: Is it used here for query string case only, isn't it?
And for version in path case I have to use SetVersionInPath filter, right?

samples/aspnetcore/SwaggerSample/Startup.cs
```diff
@@ -0,0 +1,133 @@
...
+        {
+            get
+            {
+                var basePath = PlatformServices.Default.Application.ApplicationBasePath;
+                var fileName = typeof( Startup ).GetTypeInfo().Assembly.GetName().Name + ".xml";
+                return Path.Combine( basePath, fileName );
+            }
+        }
+
+        static Info CreateInfoForApiVersion( ApiVersionDescription description )
```
xperiandri: This can be made as local function as it does not used elsewhere

samples/aspnetcore/SwaggerSample/ImplicitApiVersionParameter.cs
commonsensesoftware: I guess my comments weren't clear enough, but this particular filter handles both query string and URL path segment. Most service authors need one or the other. I wanted to show both working without creating another sample application.

In the URL path segment scenario, the only thing you need to do is assign the **Default** property.  This will fill in the current version API so that you can use the _Try It!_ feature. In the query string scenario, you have to add an implicit parameter that isn't formally declared in your controller actions. This would also be the case for versioning by header.

PR #116, opened 2017-04-24
Add header versioning sample with js inject and reuse controller actions

lurumad:
Hi,

I've added a new sample using version header (api-version) and I've added and inject a js file to send the header when you change the version selector in swagger ui.

Also, I've added two version of the same controllers but in my case I'm reusing the actions that do not change between versions. The new versioned controller only add the controller actions that have breaking changes between versions.

Regards!

### DIFF
--- a/ApiVersioningWithSamples.sln
+++ b/ApiVersioningWithSamples.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26403.3
+VisualStudioVersion = 15.0.26403.7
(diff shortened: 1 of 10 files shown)


commonsensesoftware:
After further evaluation, I can see that this example does demonstrate how to configure Swagger to support API versioning using a header-based API version. While it may be useful to some, the example demonstrates more of what _not_ to do and is more of a Swagger example than an API versioning example. For now, I don't know that it's need or should be promoted.

Thank you for taking the time though. Hopefully, you're not discourage from future contributions. Should this arise in general discussion, I may reconsider. Perhaps a more appropriate place for this would be in the wiki notes on Swagger integration.

Issue #123, opened 2017-05-03
UnsupportedApiVersion when the request is redirected from HTTP (via 302).


enihcam:
The redirect logic is
`
app.UseRewriter(new RewriteOptions().AddRedirectToHttps(StatusCodes.Status302Found, 44346));
`

Then when I accessed `http://localhost:80/subscriptions/12345?api-version=2.0`, I got:
`
{
  "Error": {
    "Code": "UnsupportedApiVersion",
    "Message": "The HTTP resource that matches the request URI 'https://localhost:44346/subscriptions/12345?api-version=2.0' does not support the API version '2.0'."
  }
}
`

But directly access to `https://localhost:44346/subscriptions/12345?api-version=2.0` worked fine.

commonsensesoftware:
This is odd. I'll have to setup a repro to test this. I would have expected the rewrite middleware to hijack the request before API versioning gets involved. There's a chance this is related to another issue I've only recently seen where the IActionSelector is called more than once, which is not expected. I'll see what I can find out. If you discover any other information, please share.

commonsensesoftware:
I didn't get a chance to verify the cause before the release candidate, but whatever the cause was, it appears to be fixed now. I set up a repro with rewriting and it worked.  Try updating to the 1.1 RC and see if that resolves your issue.  Should nothing regress, I'm planning on releasing by the end of the week.

Also keep in mind that the configuration order matters. Your configuration method should be something like:

```c#
public void Configure( IApplicationBuilder app, IHostingEnvironment env )
{
    var options = new RewriteOptions();

    // TODO: add your rewrite options

    app.UseRewriter( options );
    app.UseMvc();
    app.UseApiVersioning();
}
```

commonsensesoftware:
It seems you've figured out the issue or it was resolved in the RC release. In the event that this issue was hit by the RC1 regression, [RC2](https://www.nuget.org/packages/Microsoft.AspNetCore.Mvc.Versioning/1.1.0-rc2) has been officially published. Let me know if you run into any other issues or reopen the issue if it's unresolved. Thanks.

cleftheris:
Hi I wanted to report an additional use case that I experienced this error happening.

For instance when someone uses the `[FormatFilter]` attribute on an action. 

**the following fails upon calling the `PUT` method.**

```csharp
[Route("v{apiVersion}/products), ApiVersion("1.0")]
public class ProductsController 
{

[...]

PR #126, opened 2017-05-10
Fix Action Selector Re-entry in ASP.NET Core

### LINKED ISSUE #106: HttpPut and HttpDelete template parameter names issue
This is very similar to #72, but is not the same or it might have been the core cause of #72.

If the route templates for HttpPut and HttpDelete only have a single parameter, there is a toxic relationship between the parameter names that will cause a version error to be thrown in certain situations.

The server reports:
`Microsoft.AspNetCore.Mvc.Versioning.ApiVersionActionSelector: Information: Multiple candidate actions were found, but none matched the requested service API version '1.0'. Candidate actions: 
`

The client reports: 
```
  "Error": {
    "Code": "UnsupportedApiVersion",
    "Message": "The HTTP resource that matches the request URI 'http://localhost:27162/api/v1.0/values/5' does not support the API version '1.0'."
}
```

**Good**
```
[HttpPut("{id}")]
[HttpDelete("{id}")]

[HttpPut("{boots}")]
[HttpDelete("{id}")]
```

**Bad**
```
[HttpPut("{middle}")]
[HttpDelete("{id}")]

[HttpPut("{boots}")]
[HttpDelete("{boo}")]
```
There seems to be some kind of super/sub string issue between the names.  I did all my testing on the Put call but it turns out that a relationship between the parameter names exists that will prevent the Delete route from being matched also.

Here is an example project.  I started with a new .NET Framework 4.6.1 ASP.NET Core Web Application (.NET Framework) project.  I pruned out unnecessary methods and includes and updated all of the Microsoft.AspNetCore packages.

[https://www.dropbox.com/s/vo80iolv08fms5l/RoutingSample.zip?dl=0](url)

I have been using a combination of Postman and the PowerShell Invoke-RestMethod to call my app.
Here is my powershell script

[https://www.dropbox.com/s/eshnz68kx17bwm7/test.ps1?dl=0](url)

### LINKED ISSUE #122: Optional parameters don't work
There was controller in asp.net core 1.1 with action method with optional parameter:
```
[Route("api")]
 public abstract class BaseController {
        // omitted
        [HttpPost("[controller]/{id?}")]
        public virtual async Task<IActionResult> Post([FromBody]TDto dto, int? id, string[] include)
        {
              // omitted
         }
}
```
It worked. then versioning was added. and this pattern stops work:
```
[Route("api/v{version:apiVersion}")]
public abstract class BaseController {
```
with error

> Microsoft.AspNetCore.Mvc.Versioninig.ApiVersionActionSelector[5]  Multiple candidate actions were found, but none matched the requested service API version '1.0'. Candidate actions:

If I add constraint to route pattern for method ("[controller]/{id:int?}") it strats work again but only with supplied id value. The only solution I found was adding new action method with url template "[controller]".
Is it error or I missed something?

commonsensesoftware:
The IActionSelector can be called multiple times by the ASP.NET Core routing infrastructure, which is not currently expected. As result, some routes short-circuit early and incorrectly return HTTP status code 400. This PR refactors the API versioning route policy decisions and defers them to later in the pipeline once all candidates have been collected.

Fixes #106, Fixes #122, and possibly relates to #123.


davidl63:
Thanks!

PR #202, opened 2017-10-04
Hotfix/fix null reference exception thrown when api parameter description.model metadata is null

will-beta:


commonsensesoftware:
Looks like you ran into an issue and jump straight into fixing it. Thanks. Can you elaborate on the issue you encountered?

As I recall, the only other time I've seen an issue like this happen is when there is a route template parameter that isn't defined in any bound model. For example, when defined in a URL segment, the "API version" route parameter isn't defined as a formal parameter or part of a model anywhere. I had to add some special handling to deal with it.

I'd _prefer_ not to have `null` because that cause it's own set of other issues. However, there may not be any other reasonable alternatives.

will-beta:
@commonsensesoftware 
Just like you said. I encountered the same issue. 
There were several APIs adopting a same routing hierarchy, sharing some route parameters that may not defined as the formal functional parameters on some API methods.

commonsensesoftware:
I see. Is there any way we can provide a default model? Even without other information, any route parameter _can_ be treated as a string. I'll research this a bit as well. If we can't come up with any alternatives, then we may have to go with this solution. Thanks.

will-beta:
@commonsensesoftware 
Sorry I don’t follow what you said about “a default model”. Can you please explain it?

commonsensesoftware:
The model metadata for the parameter is unresolved in some cases, which is resulting in `null`. This only seems to happen for route parameters that are not formal parameters or part of a model. Such parameter should safely be able to be model as `string`. If we can represent the parameter with a **name** and as type `string` this would be preferable to `null`.

will-beta:
@commonsensesoftware 
Sorry for the late response! And thanks a lot for the detailed explanation!

It seems like there were some misunderstanding, and sorry again I didn't make myself clear.
I've known the reason why the NullException happened.When I asked if you could explain it, I actually wanted an example from you.

Anyway, I guess I get the idea of "a default model", which should be preparing a base class(model) with all parameters in the route and creating sub classes as request entity model for every API method. It's just I don't think this "a default model" is a flexible way to solve this problem.

commonsensesoftware:
Hey @will-beta, sorry for the hiatus, been busy with the day job lately. I'm trying to circle back around on this PR. I see that you've added a bunch of defensive things to handle `null`. I'd like to avoid null all together if possible. Do you have a repro or some sample scenario that causes this issue (e.g. nulls)? I'd like a better understanding of the scenario. Feel free to attach the *.zip here or provide a link to a source repo.

In early versions of this package, I think nulls may have been possible, but I thought I addressed all the cases where that could happen. Maybe I'm miss understanding and this is just some defensive code for things the API Explorer cannot do on its own. For example, it will handle the API version route parameter even though it's not model-bound, but that's not true for other route parameters that are not model-bound and are not known to API versioning. I'd be ok adding the checks to the examples of illustrative purposes, but these shouldn't be necessary in the API Explorer code itself.

If things are running into NRE directly in the API Explorer code, I'd like to:

1. Know what conditions or setup require that to happen
2. Know where the NRE happens in the code
3. Add tests that ensure it's properly handled when null is encountered

Thanks

commonsensesoftware:
FYI, some of the changes were stale, but I ultimately cherry-picked the remaining, relevant pieces and rolled them into the examples.  I couldn't find any remaining locations in the source where an NRE might occur. Thanks for your contribution. I'm sure it will be useful to some.

will-beta:
Hey @commonsensesoftware , truely sorry for the long long hiatus, for my crazy busy job these months!

When you debug the attached example asp.net core project [ApiVersioningTest.zip](https://github.com/Microsoft/aspnet-api-versioning/files/1614577/ApiVersioningTest.zip), a browser would pop up without the expected swagger document because of the NULL exception.
And the new `Microsoft.AspNetCore.Mvc.Versioning.ApiExplorer v2.1.0` nuget package still has this problem. 

Thanks!

commonsensesoftware:
No problem. I totally understand. Sometimes the community forgets that I have a day job and I'm running this repo by myself. ;)

Thanks for the repro. That is always help. I'll dig into to track down the issue. Thanks again!

PR #206, opened 2017-10-23
Fixed signature

Skyreach:


commonsensesoftware:
Gosh ... what a bonehead mistake. I'll be enabling StyleCop on myself to avoid these in the future. Thanks for catching and fixing this. Sadly, these fixes could be a breaking change for some, but I think everyone will agree it's the right fix. Thanks!

PR #215, opened 2017-11-21
Added ability to look for lower version implemtation

zdarovka:
When using inheritance with versioning the endpoint from lower version are not visible, this will make it work.

https://github.com/Microsoft/aspnet-api-versioning/issues/142

### DIFF
--- a/src/Microsoft.AspNet.WebApi.Versioning/Dispatcher/DirectRouteControllerSelector.cs
+++ b/src/Microsoft.AspNet.WebApi.Versioning/Dispatcher/DirectRouteControllerSelector.cs
@@ -52,6 +52,23 @@
 
             var versionedController = GetVersionedController( aggregator, requestedVersion );
 
+           var versionedController = GetVersionedController( aggregator, requestedVersion );
+
+            if ( versionedController == null )
+            {
+                if(options.LookForLowerVesions && requestedVersion.MajorVersion.HasValue)
+                {
+                    for ( int i = requestedVersion.MajorVersion.Value - 1; i > 0; i-- )
+                    {
+                        versionedController = GetVersionedController( aggregator, new ApiVersion( i, 0 ) );
+                        if ( versionedController != null )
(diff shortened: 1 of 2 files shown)


zdarovka:
Working example [InheritanceDemo.zip](https://github.com/Microsoft/aspnet-api-versioning/files/1492995/InheritanceDemo.zip)

src/Microsoft.AspNet.WebApi.Versioning/Dispatcher/DirectRouteControllerSelector.cs
```diff
@@ -52,6 +52,23 @@ internal override ControllerSelectionResult SelectController( ApiVersionControll
...
 
+           var versionedController = GetVersionedController( aggregator, requestedVersion );
+
+            if ( versionedController == null )
+            {
+                if(options.LookForLowerVesions && requestedVersion.MajorVersion.HasValue)
+                {
+                    for ( int i = requestedVersion.MajorVersion.Value - 1; i > 0; i-- )
+                    {
+                        versionedController = GetVersionedController( aggregator, new ApiVersion( i, 0 ) );
```
commonsensesoftware: Despite opting into this behavior, this is a gross over-simplification of what would be required to handle this feature in a reasonable way. The _minor_ version hasn't been considered. What happens if there is a _group version_ (aka date) or _status_ attached? The are simply not included. Furthermore, this only considers the _direct_ route scenario (aka attribute routing). It's quite possible that the one you're looking for is in the convention-based path. Mixing routing styles is a real (unfortunately) and supported scenario.

commonsensesoftware:
Since I haven't heard anything back, I'm assumming this PR is abandoned. I've queued a new issue to track interleaving across types. This combined without other options should address you ask.

zdarovka:
Hello, sorry for late reply, but actually I don't know how to make things better. This solution is working for me

PR #267, opened 2018-03-26
Support ASP.NET Core with OData

commonsensesoftware:
The implementation to support API versioning on ASP.NET Core using OData.


src/Microsoft.AspNetCore.OData.Versioning/Extensions.DependencyInjection/MetadataControllerConfiguration.cs
```diff
@@ -0,0 +1,112 @@
...
+        }
+
+        IEnumerable<ControllerModel> GetControllerModels( ApplicationModelProviderContext context )
+        {
+            Contract.Requires( context != null );
+            Contract.Ensures( Contract.Result<IEnumerable<ControllerModel>>() != null );
+
+            // TODO: is there a better way to do this? since we're already inside a pass of applying conventions, rebuild and apply
+            // the conventions without modifying the current pass to avoid unwanted side effects.
+            var provider = new DefaultApplicationModelProvider( Options );
```
philcarbone: Could you inherit from DefaultApplicationModelProvider?

src/Microsoft.AspNet.OData.Versioning/Web.OData/Routing/VersionedAttributeRoutingConvention.cs
```diff
@@ -194,37 +181,32 @@ public virtual string SelectController( ODataPath odataPath, HttpRequestMessage
...
+                    routingConventionsStore.Add( item );
+                }
+                else
+                {
+                    routeData.Values.Add( item );
                 }
             }
 
-            return null;
+            return attributeRouteData["action"] as string;
```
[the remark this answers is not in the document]
commonsensesoftware: I'm not totally sure. This is basically just following the OData source [here](https://github.com/OData/WebApi/blob/feature/netcore/src/Microsoft.AspNet.OData.Shared/Routing/Conventions/AttributeRoutingConvention.cs#L89).

src/Microsoft.AspNetCore.OData.Versioning/Extensions.DependencyInjection/MetadataControllerConfiguration.cs
commonsensesoftware: I'm leery about extending or otherwise replacing any of the **IApplicationModelProviders**. Ideally, this code needs to run **after** all of the registered **IApplicationModelProviders** have run. I'm not entirely sure how to do that - yet. If I force it to run when it's not expected, I'm not sure what will happen; maybe nothing. The stop-gap approach here was to explicitly use **DefaultApplicationModelProvider** to get what's needed without changing the configuration until a better way is found.

src/Microsoft.AspNet.OData.Versioning/Web.OData/Routing/VersionedAttributeRoutingConvention.cs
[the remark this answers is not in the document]
philcarbone: I'm just thinking about arbitrary errors, for a user-developer (just made that term up). If for some reason that was missing, they will get an Exception saying that they key is missing, but they won't really understand what do do about it. Maybe it isn't possible to tell the user-developer about the problem, if there is no case that can be identified that the key would ever by missing. Just a thought that crossed my mind as I was reading the code. 

At lease it's out here, so someone can look it up. If it ever became a problem someone can open a PR to add a message :)

src/Microsoft.AspNetCore.OData.Versioning/Extensions.DependencyInjection/MetadataControllerConfiguration.cs
philcarbone: This makes sense

cstheriot:
Any idea when these changes will be merged and a nuget package created? We will be starting to implement OData with our .Net Core apps, and the updates in the PR will be needed.

commonsensesoftware:
Weird how GitHub is showing this. Oddly, it doesn't even show the branch. Must be because of how I started.

The story up to now...

I started on this work while OData was in beta. I found a couple of issues, which they fixed, and I didn't have batching supported. In the meantime, a critical bug was discovered in the ASP.NET Core implementation by an internal team. That fix lead me down a path that pre-emptively started the work for v3.0 of all libraries. Since this is the first release, it didn't make sense to do two variations so close to each other. Plus, I just don't have that much extra bandwidth or capacity. What I end up doing was moving everything toward v3.0 and rebase this work on top of it. I think started work on the remaining pieces. I renamed the original branch to "v3", but it looks like GitHub still sees it as the original name. The commit history is clearly showing all of the new stuff.

The state of the union is that I'm hammering out the last bits of support for Swagger with OData on ASP.NET Core and the acceptance tests. There have also been some additional changes to OData since the beta that I had to account for. As soon as that is working reasonably well, I'll publish the beta. That will give me a chance to catch up on the new documentation and such. It's o-so-close. Sorry it's taken so long, but I'm up against physics. ;)

I have some unpushed changes still. If you're interested in helping, reviewing, or just testing I can push them up in a buildable state. The Swagger stuff, in particular, is in flux, but most of the rest is stabilizing.

commonsensesoftware:
@philcarbone and @cstheriot I finally have a working end-to-end build that I think is worthy of the _beta_ quality gate. There is a ton of new work and features since this PR started. Since this feature was merged into overall 3.0 change set, I'm going to close this PR out to start an official 3.0 change set. Arguably I should have done this a different way, but it is what it is. Kind of a consequence of working as a one-man show. Lesson learned for building out the next major feature set (4.0+?).

Feel free to comment, provide feedback, etc for anything new. I'm preparing the release notes and I'm hoping to get a beta release out by tomorrow. Thanks.

PR #292, opened 2018-05-11
Fix API Version Model Aggregation

commonsensesoftware:
Synchronizes dictionary access on action descriptors when the version policy is applied

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Versioning/Abstractions/ActionDescriptorExtensions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Versioning/Abstractions/ActionDescriptorExtensions.cs
@@ -24,11 +24,17 @@
             Contract.Requires( matchingActions != null );
 
             var properties = action.Properties;
+
+            if ( properties.TryGetValue( VersionPolicyIsAppliedKey, out bool applied ) && applied )
+            {
+                return;
+            }
+
             var syncRoot = properties is ICollection collection ? collection.SyncRoot : properties;
 
             lock ( syncRoot )
             {
-                if ( !properties.TryGetValue( typeof( ApiVersionModel ), out ApiVersionModel model ) || ( properties.TryGetValue( VersionPolicyIsAppliedKey, out bool applied ) && applied ) )
+                if ( !properties.TryGetValue( typeof( ApiVersionModel ), out ApiVersionModel model ) || ( properties.TryGetValue( VersionPolicyIsAppliedKey, out applied ) && applied ) )
                 {
                     return;
                 }
--- a/src/Microsoft.AspNetCore.Mvc.Versioning/Abstractions/ActionDescriptorExtensions.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Versioning/Abstractions/ActionDescriptorExtensions.cs
@@ -2,6 +2,7 @@
 {
     using ApplicationModels;
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.Diagnostics.Contracts;
     using System.Linq;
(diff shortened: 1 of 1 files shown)


src/Microsoft.AspNetCore.Mvc.Versioning/Abstractions/ActionDescriptorExtensions.cs
```diff
@@ -15,26 +16,32 @@ public static class ActionDescriptorExtensions
...
 
         internal static void AggregateAllVersions( this ActionDescriptor action, IEnumerable<ActionDescriptor> matchingActions )
         {
             Contract.Requires( action != null );
             Contract.Requires( matchingActions != null );
 
-            if ( action.VersionPolicyIsApplied() )
+            var properties = action.Properties;
+
+            if ( properties.TryGetValue( VersionPolicyIsAppliedKey, out bool applied ) && applied )
```
halter73: I still don't think this is thread-safe given concurrent callers to this method. While Dictionary does support concurrent readers, what's to prevent someone from writing during this call to TryGetValue?

commonsensesoftware: Fair question. This is the one and **only** place where the key `VersionPolicyIsAppliedKey` ever written. The key is not publicly exposed or documented. If someone else is changing this key, they've broken the intended design.

...or is there some other scenario that I'm glossing over.

rynowak: Dictionary **in general** is not thread safe for concurrent writers or one concurrent writer and any number of readers.

If you need to stash data in an action descriptor, the best way to that is by implementing

https://github.com/aspnet/Mvc/blob/dev/src/Microsoft.AspNetCore.Mvc.Abstractions/Abstractions/IActionDescriptorProvider.cs

halter73: @rynowak is right. Even methods like VersionPolicyIsApplied() need to be synchronized unless your 100% sure nothing else (e.g. AggregateAllVersions) can write to the properties dictionary at the same time.

Also, even if keys could be synchronized independently, you're basically attempting a double-check lock [which isn't advisable](http://csharpindepth.com/Articles/General/Singleton.aspx#dcl).

halter73: I'm not sure the priority or urgency of this change, but if `AggregateAllVersions` cannot be safely called in parallel as written in this PR. Even if `AggregateAllVersions` is the only thing that ever writes to this dictionary, everything that could read from it at the same time including this call to `TryGetValue`, the `VersionPolicyIsApplied` extension method, and probably more all need the same lock.

> The idea was to minimize locking as much as possible since it doesn't seem it can be removed - yet. If you feel a full lock for all read/write here is better, I can live with that.

If you go the locking route, please lock all access to the dictionary. Uncontested locks are inexpensive, and if it is highly contested, that's all the more reason to be sure this is really thread-safe.

I'm personally not concerned the right way to address this issue holistically. All I'm saying is that this PR as written isn't thread-safe and likely doesn't address the issue it was trying to solve.



, `Dictionary.TryGetValue` can and in all likelihood will throw

commonsensesoftware: Thanks for the clarity. Let me check the internal thread before abandoning this PR.

commonsensesoftware:
Abandoning this PR since the consensus to fix the systemic issue.

PR #301, opened 2018-05-22
Pre-Aggregate API Version Models

### LINKED ISSUE #306: ASP.NET Core: CPU Pegged at 100%
## Symptoms
After a non-deterministic amount of time and/or load, CPU utilization spikes and seemingly remains pegged at 100%.

## Triage
After capturing and reviewing ETL rundowns of the host process, this appears to be caused by runtime manipulation of non-concurrent dictionaries.

commonsensesoftware:
Refactor API version model aggregation to be pre-computed and supplants the need to modify action property bags at runtime. Fixes #306.

### DIFF
--- a/src/Microsoft.AspNetCore.Mvc.Versioning/Versioning/ApiVersionCollator.cs
+++ b/src/Microsoft.AspNetCore.Mvc.Versioning/Versioning/ApiVersionCollator.cs
@@ -1,0 +1,134 @@
+﻿namespace Microsoft.AspNetCore.Mvc.Versioning
+{
+    using Microsoft.AspNetCore.Mvc.Abstractions;
+    using Microsoft.AspNetCore.Mvc.Controllers;
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.Contracts;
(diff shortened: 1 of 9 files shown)


src/Microsoft.AspNetCore.Mvc.Versioning/Versioning/ApiVersionCollator.cs
```diff
@@ -0,0 +1,134 @@
...
+            {
+                var collatedModel = CollateModel( actions );
+
+                foreach ( var action in actions )
+                {
+                    var model = action.GetProperty<ApiVersionModel>();
+
+                    if ( model != null )
+                    {
+                        action.SetProperty( model.Aggregate( collatedModel ) );
```
samuelec: Is this guaranteed to be thread safe? I can see that Microsoft.AspNetCore.Mvc.Abstractions.ActionDescriptor the Properties "property" is backed by a simple Dictionary but I'm unclear of the flow here and if there is a case where there can be multiple threads accessing this (which is the problem you are trying to solve with this PR)

PR #312, opened 2018-06-20
Fix a typo in xml docs for ApiVersion

altso:


commonsensesoftware:
Doh! Good catch. Thanks.

PR #317, opened 2018-07-12
change odata route prefix to support api/version requests

cstheriot:
I was having an issue when trying to run the ODataBasicSample in the latest .Net Core with OData examples. When trying to query with `api/v1/people`, I would get a 404. I did some research and testing and made a simple change for the project to work when using the version in the path instead of in the query string.

### DIFF
--- a/samples/aspnetcore/ODataBasicSample/Controllers/OrdersController.cs
+++ b/samples/aspnetcore/ODataBasicSample/Controllers/OrdersController.cs
@@ -11,13 +11,11 @@
     public class OrdersController : ODataController
     {
         // GET ~/v1/orders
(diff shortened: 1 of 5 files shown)


commonsensesoftware:
Thanks for pointing this out. You basically had an _alpha_ build to show the basic functionality working. At the time, I was just working to make the simplest of scenarios work. Many things have been updated and/or fixed since then. The only real change was to add the additional route registration. You can see and/or test that out [here](../blob/v3/samples/aspnetcore/ODataBasicSample/Startup.cs#L43
) in the 3.0 branch.

I don't think this PR is needed any more, but thanks again for the second pair of eyes. I probably would have forgotten to update the sample.

PR #366, opened 2018-10-11
Version parsing from namespace bug fix

dlyz:
When some controller namespace contains segment starting with `v`, and the next segment contains actual version, then `VersionByNamespaceConvention` skips second segment due to dividing dot capture.

For example version parsing fails with namespace `VersioningSample.V1.Controllers`.

This fix uses regex lookahead on trailing dot.


commonsensesoftware:
I've actually gone ahead and fixed the issue separately in #377, but referenced your PR. The final pattern change just needed to be a set rather than a _non-capture_ group. I added your example to the test suite to prevent future regressions. Thanks for not only reporting the issue, but taking the time to try and fix it. It's appreciated.

PR #382, opened 2018-10-24
Support self referencing type substitution

LuukN2:
Thanks for the substitution fixes. 

However, there was still a small recursion issue when a model references itself on a property. (Or when two models reference each other, etc.)
The program would get stuck in a loop getting the structured type of the properties causing a stackoverflow exception.
Now it keeps track of what types it has already substituted in a dictionary and passes these values instead of recreating the structured type again.

Thanks again for your hard work, your project is awesome!


commonsensesoftware:
This is great! I suspected there were still some edge cases I was missing. This is a really _convenient_ feature to use, but it's not so easy to implement. Let me go through this and we'll get it merged.

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
```diff
@@ -66,6 +67,14 @@ public Type NewStructuredType( IEdmStructuredType structuredType, Type clrType,
                 var propertyType = property.PropertyType;
                 var structuredTypeRef = structuralProperty.Type;
 
+                if ( passedTypes.TryGetValue( structuredTypeRef.ToString(), out var passedType ) )
```
commonsensesoftware: Let's capture the type name so we don't have to build it two times. Something like:

```c#
var typeName = structuredTypeRef.FullName();
```

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
```diff
@@ -30,6 +30,7 @@ public sealed class DefaultModelTypeBuilder : IModelTypeBuilder
         readonly ICollection<Assembly> assemblies;
         readonly ConcurrentDictionary<ApiVersion, ModuleBuilder> modules = new ConcurrentDictionary<ApiVersion, ModuleBuilder>();
         readonly ConcurrentDictionary<ClassSignature, Type> generatedTypes = new ConcurrentDictionary<ClassSignature, Type>();
+        private readonly Dictionary<string, Type> passedTypes = new Dictionary<string, Type>();
```
commonsensesoftware: nit: let's use `visitedTypes` as it's more consistent with the pattern being used

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
```diff
@@ -66,6 +67,14 @@ public Type NewStructuredType( IEdmStructuredType structuredType, Type clrType,
                 var propertyType = property.PropertyType;
                 var structuredTypeRef = structuralProperty.Type;
 
+                if ( passedTypes.TryGetValue( structuredTypeRef.ToString(), out var passedType ) )
+                {
+                    properties.Add(new ClassProperty(property, passedType));
```
commonsensesoftware: nit: the spacing is off by my rules, but this should be flagged by code analysis and auto-formatted by the .editorconfig settings. correcting this should be as easy as formatting the document.

>Definitely let me know you don't see CA warnings and/or formatting the document doesn't fix all of these. You're not supposed to have to guess or do a lot of work to format things the way _I want_ them to be. ;)

test/Microsoft.AspNet.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
```diff
@@ -86,6 +86,29 @@ public void type_should_match_edm_with_nested_entity_substitution()
...
+
+            var context = NewContext( modelBuilder.GetEdmModel() );
+            var originalType = typeof(Company);
+
+            //act
+            var subsitutedType = originalType.SubstituteIfNecessary( context );
+
+            // assert
+            subsitutedType.GetRuntimeProperties().Should().HaveCount( 3 );
+            subsitutedType.Should().HaveProperty<int>( nameof(Company.CompanyId) );
```
commonsensesoftware: nit: spacing; format of the document should fix them all

test/Microsoft.AspNetCore.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
```diff
@@ -85,6 +85,29 @@ public void type_should_match_edm_with_nested_entity_substitution()
...
+
+        [Fact]
+        public void type_should_match_with_self_referencing_property_substitution()
+        {
+            // arrange
+            var modelBuilder = new ODataConventionModelBuilder();
+            var person = modelBuilder.EntitySet<Company>( "Companies" ).EntityType;
+
+            var context = NewContext( modelBuilder.GetEdmModel() );
+            var originalType = typeof(Company);
```
commonsensesoftware: nit: spacing; format of the document should fix them all

test/Microsoft.AspNet.OData.Versioning.ApiExplorer.Tests/Microsoft.AspNet.OData.Versioning.ApiExplorer.Tests.csproj
```diff
@@ -1,23 +1,18 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
```
commonsensesoftware: This file doesn't seem to have actually been changed. It appears that only the extra lines have been removed. It's not a huge deal, but there's not much sense in changing a file that doesn't need changing.

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
commonsensesoftware: In thinking about this, I'm pretty sure that using `string` alone for the key is not going to work because there can be a distinctly generated type by API version. This should mean that the key for the corresponding EDM type should be a combination of EDM type and API version; otherwise, it's possible that the wrong generated type will be picked.

For example, imagine that `Default.Company` is generated for API version 2.0 and then it's requested again for API version 1.0. If the models aren't exactly the same (which they probably aren't), the wrong selection is made.

There's two ways this can be handled:
1. Add a hashable struct, something like **VisitedTypeKey**
2. Use `int` as the key and add a hashing function

Option 2 might look like:

```c#
static int GetVisitedTypeKey( IEdmTypeReference type, ApiVersion apiVersion ) =>
  ( type.FullName().GetHashCode() * 397 ) ^ apiVersion.GetHashCode();
```

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
```diff
@@ -66,6 +67,14 @@ public Type NewStructuredType( IEdmStructuredType structuredType, Type clrType,
...
+                    continue;
+                }
+
+                passedTypes.Add(structuredTypeRef.ToString(), propertyType);
```
commonsensesoftware: Adding the visited type needs to be added toward the end; probably after line 102. The `propertyType` variable can be mutated after this line, which means the wrong type will be stored in the dictionary.

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
commonsensesoftware: This obviously supplanted by the struct key or hash code if you refactor

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
LuukN2: It formats for methods, but I had to manually add the formatting rules for nameof, typeof etc.
I'm using Jetbrains Rider if that means anything.

commonsensesoftware:
Thanks @LuukN2 for your contribution; really - it was useful. I'm going to close this one out because it turns out there's quite a bit more going on than just these rather basic fixes. Your change does address the _basic_ recursion problem, but it completely misses recursively emitting a self-referencing type when it's actually substituted. I realized that this was missing when incorporating your unit tests. I even had to do some research because I had no idea how to actually build a type like that. I also expanded your scenario to also include the property `List<Company> Subsidiaries { get; }`, which should be substituted as `IEnumerable<TDeclaringType> Subsidiaries { get; }` when required.

I've incorporated your bits and all the additional required changes into #384 that I already had going. The [last commit](../pull/384/commits/6699d96ec56338f421760996e1de5bd2b73e6fc2) has all of the changes relevant to this PR. I'll leave #384 open for a bit so you have some time to review and comment if you want.

Thanks again for this submission. I don't think I would have remembered about this use case otherwise.

PR #384, opened 2018-10-25
Fix OData URL Substitution

### LINKED ISSUE #365: Asp Net Core + OData + Swagger + Url Segment Versioning
Hi all,

I'm having trouble setting up versioning by URL segment. After figuring out how to add the version to the route template, I'm faced with these situations:

1. The routing works as expected. Each call is directed to the correct version of the api;
1.  Swagger does not recognize the template, and instead of registering routes like this ```api/v1/values``` it is showing like this ```api/v{version:apiVersion}/values``` which, as expected, causes a lot of troubles;
1. And finally, when trying to show the document for another version, I get and exception saying my routes are not unique.

Here is how I've setup the application:

Startup.cs
```csharp
public void ConfigureServices(IServiceCollection services)
{
    services.AddMvc()
        .SetCompatibilityVersion(CompatibilityVersion.Version_2_1);

    services.AddApiVersioning(options =>
    {
        options.ReportApiVersions = true;
        options.ApiVersionReader = new UrlSegmentApiVersionReader();
    });
    services.AddOData().EnableApiVersioning();
    services.AddODataApiExplorer(
        options =>
        {
            options.GroupNameFormat = "'v'VVV";
            options.SubstituteApiVersionInUrl = true;
        });

    services.AddSwaggerGen(a =>
    {
        var provider = services.BuildServiceProvider().GetRequiredService<IApiVersionDescriptionProvider>();

        foreach (var description in provider.ApiVersionDescriptions)
        {
            var info = new Info
            {
                Title = "My API Title",
                Version = description.ApiVersion.ToString(),
                Description = "My API Description"
            };

            if (description.IsDeprecated)
                info.Description += " NOTE: This API has been deprecated";

            a.SwaggerDoc(description.GroupName, info);
        }

        a.ParameterFilter<SwaggerDefaultValues>();

        var filePath = Path.Combine(PlatformServices.Default.Application.ApplicationBasePath, $"{PlatformServices.Default.Application.ApplicationName}.xml");
        a.IncludeXmlComments(filePath);

        a.DescribeAllEnumsAsStrings();
    });
}

public void Configure(IApplicationBuilder app, IHostingEnvironment env, VersionedODataModelBuilder versionedModelBuilder, IApiVersionDescriptionProvider versionDescriptionProvider)
{
    app.UseMvc(routeBuilder =>
    {
        routeBuilder.Select().Expand().Filter().OrderBy().Count();
        routeBuilder.MapVersionedODataRoutes("odata", "api/v{v:apiVersion}", versionedModelBuilder.GetEdmModels());
    });
    app.UseSwagger();
    app.UseSwaggerUI(a =>
    {
        // build a swagger endpoint for each discovered API version
        foreach (var description in versionDescriptionProvider.ApiVersionDescriptions)
        {
            a.SwaggerEndpoint($"/swagger/{description.GroupName}/swagger.json", description.GroupName);
        }
    });
}
```

SwaggerDefaultValues.cs
```csharp
public class SwaggerDefaultValues : IParameterFilter
{
    public void Apply(IParameter parameter, ParameterFilterContext context)
    {
        if (!(parameter is NonBodyParameter nonBodyParameter))
            return;

        if (nonBodyParameter.Description == null)
        {
            nonBodyParameter.Description = context.ApiParameterDescription.ModelMetadata?.Description;
        }

        if (context.ApiParameterDescription.RouteInfo == null)
            return;

        if (nonBodyParameter.Default == null)
        {
            nonBodyParameter.Default = context.ApiParameterDescription.RouteInfo.DefaultValue;
        }

        parameter.Required |= !context.ApiParameterDescription.RouteInfo.IsOptional;
    }
}
```

So, am I doing something wrong? Is this scenario supported?

Thanks in advance.

commonsensesoftware:
Fixes the parsing and substitution of route prefixes for the OData API Explorer. Also enables the optional inclusion of API version parameters for version-neutral actions (normally excluded).

Fixes #365

### DIFF
--- a/build/file-version.targets
+++ b/build/file-version.targets
@@ -7,7 +7,7 @@
     <MajorAndMinorVersion>$(AssemblyVersion.Split(`.`)[0]).$(AssemblyVersion.Split(`.`)[1])</MajorAndMinorVersion>
     <DaylightSavingTime>$([System.DateTime]::Now.IsDaylightSavingTime())</DaylightSavingTime>
     <FileBuildNumber>$([System.DateTime]::Today.Subtract($([System.DateTime]::Parse("1/1/2000"))).ToString("%d"))</FileBuildNumber>
-    <FileBuildRevision Condition=" '$(DaylightSavingTime)'=='True' " >$([System.Convert]::ToInt32($([MSBuild]::Divide($([System.DateTime]::Now.TimeOfDay.Subtract($([System.TimeSpan]::FromHours(1.0))).TotalSeconds),2))))</FileBuildRevision>
+    <FileBuildRevision Condition=" '$(DaylightSavingTime)'=='True' " >$([System.Convert]::ToInt32($([MSBuild]::Divide($([System.DateTime]::Now.Subtract($([System.TimeSpan]::FromHours(1.0))).TimeOfDay.TotalSeconds),2))))</FileBuildRevision>
     <FileBuildRevision Condition=" '$(DaylightSavingTime)'=='False' " >$([System.Convert]::ToInt32($([MSBuild]::Divide($([System.DateTime]::Now.TimeOfDay.TotalSeconds),2))))</FileBuildRevision>
     <FileVersion>$(MajorAndMinorVersion).$(FileBuildNumber).$(FileBuildRevision)</FileVersion>
   </PropertyGroup>
--- a/samples/webapi/AdvancedODataWebApiSample/Startup.cs
+++ b/samples/webapi/AdvancedODataWebApiSample/Startup.cs
@@ -2,12 +2,14 @@
 
 namespace Microsoft.Examples
 {
-    using Configuration;
     using global::Owin;
+    using Microsoft.AspNet.OData;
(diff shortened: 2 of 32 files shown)


LuukN2:
This should fix self-references.

However I got an algorithm working that works for both self-referencing types and back-references.
The downside however is that when there is a self-reference or back-reference present you cannot know if the referenced type matches the clrtype yet. This means these types will always be dynamically generated.

It keeps track of the types that couldn't be finished because of properties that still have types being built for them and adds them to a list of properties that have dependencies on those types.

When a type is being built it checks this list for properties that reference it. It adds a property on that type that references the TypeBuilder. If this depletes the list of dependencies for that type the type can be built and is added to the generated types.

I have commited very rough changes on my branch if you want to check them out, I'll rewrite it based on these changes, polish it and open a new pull request later.

I have a lot of time available next week and there's a few more possible edge cases that I'll make repros/pull requests for next week.

commonsensesoftware:
Sounds good. I appreciate the help. :)

PR #390, opened 2018-10-29
Support substitution with dependencies

LuukN2:
Allow the substitution of structured types that have references to types that are still being built.
This supports self-references but also adds support for backreferences or circular references.

Removed the NewStructuredTypeOrSelf methods because the dependency resolver will handle this now.

Note: I removed the type_should_match_with_self_referencing_property_substition test because with the current implementation if there is a self-reference or back reference the type will always be dynamically built.

### DIFF
--- a/src/Common.OData.ApiExplorer/AspNet.OData/ClassProperty.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/ClassProperty.cs
@@ -38,17 +38,8 @@
 
         public override int GetHashCode() => ( Name.GetHashCode() * 397 ) ^ type.GetHashCode();
 
-        public Type GetType( Type declaringType )
+        public Type GetPropertyType()
         {
-            Contract.Requires( declaringType != null );
-            Contract.Ensures( Contract.Result<Type>() != null );
-
-            if ( type.IsGenericType )
-            {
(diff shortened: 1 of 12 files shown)


commonsensesoftware:
Has all use of **DeclaringType** been removed? It looks like it. If so, then we might as well get rid of it as well as the `TypeExtensions.IsDeclaringType` extension method. It would seem your approach doesn't require this. That's fine by me. I just don't want to keep _dead_ code. ;)

Beyond that, eveything else looks good. I see my CI build for third-party PRs is working properly again too. Let's tie off this tiny change and I think it's good to merge.

commonsensesoftware:
Not sure if it gave the notice. It looks like that last change broke a few tests. I suspect these were tests that were affected by the last round of changes. I don't know that there's a way for you to see them since I have to host with internal CI build servers. =/  There were 9 failures, all within **DefaultModelTypeBuilderTest**.

LuukN2:
Fixed it, had to change how the type was gotten from the ClassProperty.
Didn't catch that some tests failed before, my bad. The tests succeed locally now so it should work.

commonsensesoftware:
No problem. That's why tests exist. We're all faulible. This is great work. Thanks.

PR #392, opened 2018-10-31
Add support for action parameters with collection parameters

LuukN2:
Just a little thing I noticed while testing. ODataActionParameters with collection parameters would cause the type in the ClassProperty to be null. This caused an exception when trying to generate the type for the parameters.

I also fixed some misplaced using directives from my previous PR.

### DIFF
--- a/src/Common.OData.ApiExplorer/AspNet.OData/ClassProperty.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/ClassProperty.cs
@@ -30,7 +30,18 @@
             Contract.Requires( parameter != null );
 
             Name = parameter.Name;
-            type = parameter.Type.Definition.GetClrType( assemblies );
+
+            if ( parameter.Type.IsCollection() )
+            {
+                var collectionType = parameter.Type.AsCollection();
+                var elementType = collectionType.ElementType().Definition.GetClrType( assemblies );
+                type = typeof( IEnumerable<> ).MakeGenericType( elementType );
+            }
+            else
+            {
+                type = parameter.Type.Definition.GetClrType( assemblies );
+            }
+
             Attributes = AttributesFromOperationParameter( parameter );
         }
 
--- a/src/Common.OData.ApiExplorer/AspNet.OData/ClassProperty.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/ClassProperty.cs
@@ -11,7 +11,7 @@
 
     struct ClassProperty
     {
-        readonly Type type;
+        internal readonly Type Type;
         internal readonly string Name;
 
         internal ClassProperty( PropertyInfo clrProperty, Type propertyType )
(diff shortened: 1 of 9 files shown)


src/Common.OData.ApiExplorer/AspNet.OData/ClassProperty.cs
```diff
@@ -30,7 +30,18 @@ internal ClassProperty( IEnumerable<Assembly> assemblies, IEdmOperationParameter
             Contract.Requires( parameter != null );
 
             Name = parameter.Name;
-            type = parameter.Type.Definition.GetClrType( assemblies );
+
+            if ( parameter.Type.IsCollection() )
+            {
+                var collectionType = parameter.Type.AsCollection();
+                var elementType = collectionType.ElementType().Definition.GetClrType( assemblies );
```
commonsensesoftware: Don't we still have to handle possible substitution here if the element type is an _Entity Type_ or _Complex Type_? The element will be a type reference, but the registered reference _might_ have filtered members in this version, no?

LuukN2: Yeah you're right.

I had to change the interface so a serviceprovider required for making the substitutioncontext is passed. I also added a new test for action parameters with types that should be substituted.

src/Common.OData.ApiExplorer/AspNet.OData/ClassProperty.cs
```diff
@@ -24,13 +24,29 @@ internal ClassProperty( PropertyInfo clrProperty, Type propertyType )
             Attributes = AttributesFromProperty( clrProperty );
         }
 
-        internal ClassProperty( IEnumerable<Assembly> assemblies, IEdmOperationParameter parameter )
+        internal ClassProperty( IServiceProvider services, IEnumerable<Assembly> assemblies, IEdmOperationParameter parameter, IModelTypeBuilder typeBuilder )
         {
```
commonsensesoftware: Need to verify that `services` and `typeBuilder` are not `null` here.

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
```diff
@@ -171,14 +171,14 @@ public Type NewStructuredType( IEdmStructuredType structuredType, Type clrType,
         }
 
         /// <inheritdoc />
-        public Type NewActionParameters( IEdmAction action, ApiVersion apiVersion )
+        public Type NewActionParameters( IServiceProvider services, IEdmAction action, ApiVersion apiVersion )
         {
```
commonsensesoftware: `services` is not allowed to be `null` here.

src/Common.OData.ApiExplorer/AspNet.OData/ClassProperty.cs
commonsensesoftware: Also, do you mind throwing in one more small tweak here? `GetPropertyType` was original a read-only field, then I changed it to a method, but now we don't need the method at all any more. Can you revert the field declaration back to:

```c#
internal readonly Type Type;
```

and dump the `GetPropertyType` method altogether?

src/Common.OData.ApiExplorer/AspNet.OData/IModelTypeBuilder.cs
```diff
@@ -29,14 +29,15 @@ public interface IModelTypeBuilder
         Type NewStructuredType( IEdmStructuredType structuredType, Type clrType, ApiVersion apiVersion );
 
         /// <summary>
-        /// Creates an returns a stronly-typed definition for OData action parameters.
```
commonsensesoftware: Nice catch. :)

commonsensesoftware:
Nicely done. Rolling it in.

PR #394, opened 2018-11-07
Type building fixes

LuukN2:
Various fixes for the type builder.

- Attributes that take **params** parameters would cause a cast exception. The expected type was array, but getting the constructor arguments converts params to a list of CustomAttributeTypedArgument.
Fixed by calling ToArray on the parameter.

- Some types that were finished still had dependencies on them. Fixed by checking if the type the property is dependent on is present in the dictionary of generated types.

- Building ODataActionParameters types caused a name conflict when actions with the same namespace and name were defined in other controllers. Fixed by passing the controller name and prefixing the typename with it.

- Prevent null pointer exception when no ODataActionParameters are found on actions. Having this parameter is not required for actions that take no parameters.

- Added support for substituting SingleResult, which is an IQuerable with one element.

Still needs test cases but with these fixes I can use the api explorer without any exceptions on the big OData project I'm working with. If everything seems ok, I'll make some test cases.
I can make some repros if needed, let me know.

It still takes around 8 minutes to generate the api descriptions, and I'll be looking in to that after this has been resolved.

### DIFF
--- a/src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
@@ -173,6 +173,7 @@
         /// <inheritdoc />
         public Type NewActionParameters( IServiceProvider services, IEdmAction action, ApiVersion apiVersion )
         {
+            Arg.NotNull( services, nameof( services ) );
             Arg.NotNull( action, nameof( action ) );
             Arg.NotNull( apiVersion, nameof( apiVersion ) );
             Contract.Ensures( Contract.Result<Type>() != null );
@@ -202,7 +203,7 @@
 
             foreach ( var property in @class.Properties )
             {
-                var type = property.GetPropertyType();
+                var type = property.Type;
                 var name = property.Name;
                 var propertyBuilder = AddProperty( typeBuilder, type, name );
 
--- a/src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
@@ -171,14 +171,14 @@
         }
 
         /// <inheritdoc />
-        public Type NewActionParameters( IEdmAction action, ApiVersion apiVersion )
+        public Type NewActionParameters( IServiceProvider services, IEdmAction action, ApiVersion apiVersion )
         {
             Arg.NotNull( action, nameof( action ) );
             Arg.NotNull( apiVersion, nameof( apiVersion ) );
             Contract.Ensures( Contract.Result<Type>() != null );
 
             var name = action.FullName() + "Parameters";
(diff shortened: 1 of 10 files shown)


commonsensesoftware:
This is great. It looks to be in the right direction to me.

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
```diff
@@ -171,14 +171,15 @@ public Type NewStructuredType( IEdmStructuredType structuredType, Type clrType,
...
+        public Type NewActionParameters( IServiceProvider services, IEdmAction action, ApiVersion apiVersion, string controllerName )
         {
             Arg.NotNull( services, nameof( services ) );
             Arg.NotNull( action, nameof( action ) );
             Arg.NotNull( apiVersion, nameof( apiVersion ) );
+            Arg.NotNull( controllerName, nameof(controllerName) );
             Contract.Ensures( Contract.Result<Type>() != null );
 
-            var name = action.FullName() + "Parameters";
+            var name = controllerName + "." + action.FullName() + "Parameters";
```
commonsensesoftware: I can't remember now, but does this handle whether different versions of the action exist by API version? I _think_ it's still possible for the `controllerName` and `action.FullName()` to be the same. Simply calling to `apiVersion.ToString()` and adding that to the name probably won't work because it will produce an invalid name. Maybe `apiVersion.GetHashCode().ToString()` somewhere in the mix? Does this even matter? Isn't it handled by the types being generated in separate assemblies per API version? I guess is just covering the case where two different controllers define actions with the same name? Just double-checking.

LuukN2: The problem I was having is duplicate type names within the same assembly. This is only per api version since a new assembly is created per api version.

This covers the case where two actions with the same name would be defined in the same api version in different controllers.

commonsensesoftware:
Just some quick items:

1. Is this the final change set or is there more?
2. Somehow there is branch conflict that I'm not following because the last commits on the remote **master** have only come from you actually.
3. For the performance stuff, CLR type lookup can definitely be improved and that can be achieved by checking for and updating the **ClrTypeAnnotation** on an EDM type as necessary.

LuukN2:
Yeah, these should be the final changes for this PR.

I'm not sure either, it said I had conflicts even though my branch was up-to-date.

Right now I have a static dictionary which increases performance drastically, thought it takes up some memory. The profiler also shows that 98% of time spent is in GetClrType.

commonsensesoftware:
Sounds good. I'm trying to get Beta 2 out by EOD. Got some other folks waiting on me. I'll see if I can't figure out what the branch issue here is so we can get it in. The performance stuff can come in the RC. Have a look at: https://github.com/OData/WebApi/blob/feature/netcore/src/Microsoft.AspNet.OData.Shared/Formatter/EdmLibHelpers.cs. Specifically, you probably want to use the annotation API to lookup and set the built-in ClrTypeAnnotation rather than maintain a separate dictionary.

PR #398, opened 2018-11-12
Use clrtypeannotation to get clrtypes + underlying property field fix

LuukN2:
Instead of iterating through assemblies use the ClrTypeAnnotation from Microsoft.AspNet.OData to get clr types.

All IEdmSchemaTypes should be decorated with this property so this should prevent iteration over assemblies to find the corresponding clr type. This makes performance a lot better, it only takes around 10 seconds to generate the api description now.

Changed a lot of methods that were passing around the assemblies for only this, I think I removed all unused lists of assemblies.

Also includes a minor fix needed for deserialization of substituted types. The underlying property field had the exact same name as the property name which caused some issues when trying to generate JSON.

### DIFF
--- a/src/Common.OData.ApiExplorer/OData.Edm/EdmExtensions.cs
+++ b/src/Common.OData.ApiExplorer/OData.Edm/EdmExtensions.cs
@@ -17,15 +17,10 @@
 
     static class EdmExtensions
     {
-#if WEBAPI
-        internal static Type GetClrType( this IEdmType edmType, IAssembliesResolver assembliesResolver ) =>
-            edmType.GetClrType( assembliesResolver.GetAssemblies() );
-#endif
-
-        internal static Type GetClrType( this IEdmType edmType, IEnumerable<Assembly> assemblies )
+        internal static Type GetClrType( this IEdmType edmType, IEdmModel edmModel )
         {
             Contract.Requires( edmType != null );
-            Contract.Requires( assemblies != null );
+            Contract.Requires(edmModel != null );
 
             if ( !( edmType is IEdmSchemaType schemaType ) )
             {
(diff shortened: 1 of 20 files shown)


src/Common.OData.ApiExplorer/OData.Edm/EdmExtensions.cs
```diff
@@ -40,17 +35,11 @@ internal static Type GetClrType( this IEdmType edmType, IEnumerable<Assembly> as
                 return type;
             }
 
-            using ( var matchingTypes = GetMatchingTypes( typeName, assemblies ).GetEnumerator() )
+            var element = (IEdmSchemaType) edmType;
+            var annotationValue = edmModel.GetAnnotationValue<ClrTypeAnnotation>(element);
+            if ( annotationValue != null )
```
commonsensesoftware: I didn't see any calls where the **ClrTypeAnnotation** is set. How or when does that happen? Shouldn't that happen right after we build a type? Maybe we were already doing that?

There should almost no case where this method returns `null`. As I recall, the built-in OData code does this lazily as it deems necessary. The tests pass and the performance has improved so I trying to figure out what I'm missing. I also wanted to double-checking that we didn't miss an edge case; otherwise, the changes look solid.

LuukN2: The ClrTypeAnnotation is set for all types when the ODataModel is being built.

https://github.com/OData/WebApi/blob/90a24d55715b0afe54404a34bf2df4e0c870de14/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs#L431-L438

Here the AddClrTypeAnnotations method is called to decorate all EdmTypes with the corresponding clrtype.

https://github.com/OData/WebApi/blob/90a24d55715b0afe54404a34bf2df4e0c870de14/src/Microsoft.AspNet.OData.Shared/Builder/EdmModelHelperMethods.cs#L494-L503

The EdmTypeMap is made from the dictionary that is kept by the model builder. This dictionary is made from the clr types corresponding to the IEdmTypeConfigurations.

https://github.com/OData/WebApi/blob/d02bc61ea7b31ada1e54abbeebbecb3c5df0e3ac/src/Microsoft.AspNet.OData.Shared/Builder/IEdmTypeConfiguration.cs#L13-L19

As far as I know the ClrTypeAnnotation should always be set so there is no need to check for a missing annotation. Good thing you made me double check, I found a different bug that I can still commit to this PR now :P

PR #425, opened 2019-01-02
Fixed type builder not keeping track of clrtypes that match the edm

LuukN2:
This is was a small oversight where types that have properties referencing other types that matched the edm would never resolve their property. This caused the type to never be resolved and throw an exception when being explored.

### DIFF
--- a/src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
@@ -132,7 +132,7 @@
 
             if ( clrTypeMatchesEdmType )
             {
-                return clrType;
+                return generatedEdmTypes.GetOrAdd( typeKey, clrType.GetTypeInfo() );
             }
 
             var signature = new ClassSignature( clrType, properties, apiVersion );
--- a/src/Microsoft.AspNet.OData.Versioning.ApiExplorer/ReleaseNotes.txt
+++ b/src/Microsoft.AspNet.OData.Versioning.ApiExplorer/ReleaseNotes.txt
@@ -1,2 +1,1 @@
-﻿Fixed query options for non-action POST
-Fix tracking CLR types that match the EDM (#425)
+﻿Fix tracking CLR types that match the EDM (#425)
--- a/src/Microsoft.AspNet.OData.Versioning.ApiExplorer/ReleaseNotes.txt
+++ b/src/Microsoft.AspNet.OData.Versioning.ApiExplorer/ReleaseNotes.txt
@@ -1,1 +1,2 @@
 ﻿Fixed query options for non-action POST
+Fix tracking CLR types that match the EDM (#425)
--- a/test/Microsoft.AspNet.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
+++ b/test/Microsoft.AspNet.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
@@ -8,6 +8,7 @@
     using System.Collections.Generic;
     using System.Linq;
     using System.Reflection;
+    using System.Reflection.Emit;
     using Microsoft.Extensions.DependencyInjection;
     using Xunit;
 
@@ -375,6 +376,29 @@
 
             Assert.Equal( "Manager", attributeWithParams.AllowedRoles[0] );
             Assert.Equal( "Employer", attributeWithParams.AllowedRoles[1] );
+        }
+
+        [Fact]
+        public void substitute_should_resolve_types_that_reference_a_model_that_match_the_edm()
+        {
+            // arrange
+            var modelBuilder = new ODataConventionModelBuilder();
+
+            var shipment = modelBuilder.EntitySet<Shipment>( "Shipments" ).EntityType;
+            shipment.Ignore( s => s.ShippedOn );
+            var originalType = typeof(Shipment);
+
+            modelBuilder.EntitySet<Address>("Addresses");
+            var addressType = typeof(Address);
+
+            var context = NewContext( modelBuilder.GetEdmModel() );
+
+            // act
+            addressType.SubstituteIfNecessary( context );
+            var substitutionType = originalType.SubstituteIfNecessary( context );
+
+            // assert
+            Assert.False( substitutionType is TypeBuilder );
         }
 
         public static IEnumerable<object[]> SubstitutionNotRequiredData
--- a/test/Microsoft.AspNet.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
+++ b/test/Microsoft.AspNet.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
@@ -252,7 +252,7 @@
             // arrange
             var modelBuilder = new ODataConventionModelBuilder();
             var contact = modelBuilder.EntitySet<Contact>( "Contacts" ).EntityType;
-            contact.Ignore( c => c.Email  );
+            contact.Ignore( c => c.Email );
             var action = contact.Action( "PlanInterview" );
 
             action.Parameter<DateTime>( "when" );
@@ -315,7 +315,7 @@
         [Fact]
         public void substitute_should_generate_types_for_actions_with_the_same_name_in_different_controllers()
         {
-            // arrange 
+            // arrange
             var modelBuilder = new ODataConventionModelBuilder();
             var contact = modelBuilder.EntitySet<Contact>( "Contacts" ).EntityType;
             var action = contact.Action( "PlanMeeting" );
@@ -330,7 +330,7 @@
             action.Parameter<DateTime>( "when" );
             action.CollectionParameter<Employee>( "attendees" );
             action.Parameter<string>( "project" );
-            
+
             var context = NewContext( modelBuilder.GetEdmModel() );
             var model = context.Model;
 
@@ -338,7 +338,7 @@
             var operations = model.FindDeclaredOperations( qualifiedName ).Select( o => (IEdmAction) o ).ToArray();
             var services = new ServiceCollection();
             services.AddSingleton( model );
-            
+
             // act
             var contactActionType = context.ModelTypeBuilder.NewActionParameters( services.BuildServiceProvider(), operations[0], ApiVersion.Default, contact.Name );
             var employeesActionType = context.ModelTypeBuilder.NewActionParameters( services.BuildServiceProvider(), operations[1], ApiVersion.Default, employee.Name );
@@ -364,7 +364,7 @@
             var employee = modelBuilder.EntitySet<Employee>( "Employees" ).EntityType;
             employee.Ignore( e => e.FirstName );
             var originalType = typeof( Employee );
-            
+
             var context = NewContext( modelBuilder.GetEdmModel() );
 
             // act
@@ -386,10 +386,10 @@
 
             var shipment = modelBuilder.EntitySet<Shipment>( "Shipments" ).EntityType;
             shipment.Ignore( s => s.ShippedOn );
-            var originalType = typeof(Shipment);
+            var originalType = typeof( Shipment );
 
-            modelBuilder.EntitySet<Address>("Addresses");
-            var addressType = typeof(Address);
+            modelBuilder.EntitySet<Address>( "Addresses" );
+            var addressType = typeof( Address );
 
             var context = NewContext( modelBuilder.GetEdmModel() );
 
--- a/test/Microsoft.AspNetCore.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
+++ b/test/Microsoft.AspNetCore.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
@@ -252,7 +252,7 @@
             // arrange
             var modelBuilder = new ODataConventionModelBuilder();
             var contact = modelBuilder.EntitySet<Contact>( "Contacts" ).EntityType;
-            contact.Ignore( c => c.Email  );
+            contact.Ignore( c => c.Email );
             var action = contact.Action( "PlanInterview" );
 
             action.Parameter<DateTime>( "when" );
@@ -315,7 +315,7 @@
         [Fact]
         public void substitute_should_generate_types_for_actions_with_the_same_name_in_different_controllers()
         {
-            // arrange 
+            // arrange
             var modelBuilder = new ODataConventionModelBuilder();
             var contact = modelBuilder.EntitySet<Contact>( "Contacts" ).EntityType;
             var action = contact.Action( "PlanMeeting" );
@@ -330,7 +330,7 @@
             action.Parameter<DateTime>( "when" );
             action.CollectionParameter<Employee>( "attendees" );
             action.Parameter<string>( "project" );
-            
+
             var context = NewContext( modelBuilder.GetEdmModel() );
             var model = context.Model;
 
@@ -338,7 +338,7 @@
             var operations = model.FindDeclaredOperations( qualifiedName ).Select( o => (IEdmAction) o ).ToArray();
             var services = new ServiceCollection();
             services.AddSingleton( model );
-            
+
             // act
             var contactActionType = context.ModelTypeBuilder.NewActionParameters( services.BuildServiceProvider(), operations[0], ApiVersion.Default, contact.Name );
             var employeesActionType = context.ModelTypeBuilder.NewActionParameters( services.BuildServiceProvider(), operations[1], ApiVersion.Default, employee.Name );
@@ -354,6 +354,28 @@
             employeesActionType.Should().HaveProperty<DateTimeOffset>( "when" );
             Assert.NotNull(employeesActionType.GetRuntimeProperty( "attendees" ));
             employeesActionType.Should().HaveProperty<string>( "project" );
+        }
+
+        [Fact]
+        public void substitute_should_get_attributes_from_property_that_has_attributes_that_takes_params()
+        {
+            // arrange
+            var modelBuilder = new ODataConventionModelBuilder();
+            var employee = modelBuilder.EntitySet<Employee>( "Employees" ).EntityType;
+            employee.Ignore( e => e.FirstName );
+            var originalType = typeof( Employee );
+
+            var context = NewContext( modelBuilder.GetEdmModel() );
+
+            // act
+            var substitutionType = originalType.SubstituteIfNecessary( context );
+
+            // assert
+            var property = substitutionType.GetRuntimeProperty( "Salary" );
+            var attributeWithParams = property.GetCustomAttribute<AllowedRolesAttribute>();
+
+            Assert.Equal( "Manager", attributeWithParams.AllowedRoles[0] );
+            Assert.Equal( "Employer", attributeWithParams.AllowedRoles[1] );
         }
 
         [Fact]
@@ -377,28 +399,6 @@
 
             // assert
             Assert.False( substitutionType is TypeBuilder );
-        }
-
-        [Fact]
-        public void substitute_should_get_attributes_from_property_that_has_attributes_that_takes_params()
-        {
-            // arrange
-            var modelBuilder = new ODataConventionModelBuilder();
-            var employee = modelBuilder.EntitySet<Employee>( "Employees" ).EntityType;
-            employee.Ignore( e => e.FirstName );
-            var originalType = typeof( Employee );
-            
-            var context = NewContext( modelBuilder.GetEdmModel() );
-
-            // act
-            var substitutionType = originalType.SubstituteIfNecessary( context );
-
-            // assert
-            var property = substitutionType.GetRuntimeProperty( "Salary" );
-            var attributeWithParams = property.GetCustomAttribute<AllowedRolesAttribute>();
-
-            Assert.Equal( "Manager", attributeWithParams.AllowedRoles[0] );
-            Assert.Equal( "Employer", attributeWithParams.AllowedRoles[1] );
         }
 
         public static IEnumerable<object[]> SubstitutionNotRequiredData
--- a/test/Microsoft.AspNetCore.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
+++ b/test/Microsoft.AspNetCore.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
@@ -8,6 +8,7 @@
     using System.Collections.Generic;
     using System.Linq;
     using System.Reflection;
+    using System.Reflection.Emit;
     using Microsoft.Extensions.DependencyInjection;
     using Xunit;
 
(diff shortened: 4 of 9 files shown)


src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
```diff
@@ -132,7 +132,7 @@ public Type NewStructuredType( IEdmStructuredType structuredType, Type clrType,
 
             if ( clrTypeMatchesEdmType )
             {
-                return clrType;
+                return generatedEdmTypes.GetOrAdd( typeKey, clrType.GetTypeInfo() );
```
commonsensesoftware: Just to confirm. What we're really saying here is that _"Once you go down that dark path, forever will it control your destiny..."_ In other words, once the type substitution path has been entered, any and all child types within scope will also be substituted, even if they match the EDM type.

I suppose that makes sense and is simpler. It would be hard to track and then perform substitution if you discover it isn't required until 3+ levels down in a hierarchy. Performing all the work in a single pass is far easier to rationalize and will likely perform better in most scenarios.

It's not 100% clear to me what the performance impact is of emitting new types, but I believe this change will result in substitution occurring in 80-90% of the cases, even when the CLR type matches the EDM type. I assume you discovered this working on your own (presumably large) API set and didn't observe any noticeable performance difference. The bar should be that the cost to generate a new type should be less than or equal to the time that it takes to compare the equality of two object graphs.

commonsensesoftware:
Is this just an edge you discovered that _should_ be fixed or is this a bug that's acutally blocking you? That determines if this fix needs to go out now or in the next round of fixes or release, whichever comes first.

If it's blocking you, then we should add the following changes to the PR. In the projects:

* Microsoft.AspNet.OData.Versioning.ApiExplorer
* Microsoft.AspNetCore.OData.Versioning.ApiExplorer

1. Bump the patch version
2. Update the release notes

To update the release notes, simply update the **ReleaseNotes.txt** file in the respective project. Use only one line per release note and strive to keep the text under 50 characters so it doesn't wrap when used by NuGet. I also include the associated issue if there is one (a PR is navigable as an issue).  For example, something like:

`Fix tracking CLR types that match the EDM (#425)`

The build process will handle everything else. Each release note line is turned into a bullet item for NuGet.

src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
Discjoggy: Thank you very much @LuukN2 !
This simple line of code fixed an exception which occured the first time in version 3.0.0. The API was stil working but the Swagger documentation caused always an exception. So this line of the `StructuredTypeResolver.cs` class..

```cs
var original = type.GetCustomAttribute<OriginalTypeAttribute>( inherit: false );
```

.. throwed the following exception:

```txt
System.NotSupportedException: "The invoked member is not supported before the type is created."

    at System.Reflection.Emit.TypeBuilder.GetCustomAttributes(Type attributeType, Boolean inherit)
    at System.Attribute.GetCustomAttributes(MemberInfo element, Type type, Boolean inherit)
    at System.Attribute.GetCustomAttribute(MemberInfo element, Type attributeType, Boolean inherit)
    at System.Reflection.CustomAttributeExtensions.GetCustomAttribute[T](MemberInfo element, Boolean inherit)
    at Microsoft.AspNet.OData.StructuredTypeResolver.GetStructuredType(Type type) in C:\Users\tsibera\source\repos\aspnet-api-versioning\src\Common.OData.ApiExplorer\AspNet.OData\StructuredTypeResolver.cs:line 24
    at Microsoft.AspNet.OData.Builder.ODataAttributeVisitor.Visit(ApiDescription apiDescription) in C:\Users\tsibera\source\repos\aspnet-api-versioning\src\Common.OData.ApiExplorer\AspNet.OData\Builder\ODataAttributeVisitor.cs:line 63
    at Microsoft.AspNet.OData.Builder.ODataValidationSettingsConvention.ApplyTo(ApiDescription apiDescription) in C:\Users\tsibera\source\repos\aspnet-api-versioning\src\Microsoft.AspNetCore.OData.Versioning.ApiExplorer\AspNet.OData\Builder\ODataValidationSettingsConvention.cs:line 38
    at Microsoft.AspNet.OData.Builder.ODataControllerQueryOptionConvention.ApplyTo(ApiDescription apiDescription) in C:\Users\tsibera\source\repos\aspnet-api-versioning\src\Common.OData.ApiExplorer\AspNet.OData\Builder\ODataControllerQueryOptionConvention.cs:line 45
    at Microsoft.AspNet.OData.Builder.ODataQueryOptionsConventionBuilder.ApplyTo(IEnumerable`1 apiDescriptions, ODataQueryOptionSettings queryOptionSettings) in C:\Users\tsibera\source\repos\aspnet-api-versioning\src\Common.OData.ApiExplorer\AspNet.OData\Builder\ODataQueryOptionsConventionBuilder.cs:line 156
    at Microsoft.AspNetCore.Mvc.ApiExplorer.ODataApiDescriptionProvider.ExploreQueryOptions(IEnumerable`1 apiDescriptions, ODataUriResolver uriResolver) in C:\Users\tsibera\source\repos\aspnet-api-versioning\src\Microsoft.AspNetCore.OData.Versioning.ApiExplorer\AspNetCore.Mvc.ApiExplorer\ODataApiDescriptionProvider.cs:line 278
    at Microsoft.AspNetCore.Mvc.ApiExplorer.ODataApiDescriptionProvider.OnProvidersExecuted(ApiDescriptionProviderContext context) in C:\Users\tsibera\source\repos\aspnet-api-versioning\src\Microsoft.AspNetCore.OData.Versioning.ApiExplorer\AspNetCore.Mvc.ApiExplorer\ODataApiDescriptionProvider.cs:line 207
    at Microsoft.AspNetCore.Mvc.ApiExplorer.ApiDescriptionGroupCollectionProvider.GetCollection(ActionDescriptorCollection actionDescriptors)
    at Microsoft.AspNetCore.Mvc.ApiExplorer.ApiDescriptionGroupCollectionProvider.get_ApiDescriptionGroups()
    at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwagger(String documentName, String host, String basePath, String[] schemes)
    at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvider)
```

I am not sure if this the excat same error which @LuukN2 has described. If so, may this can help to identify if this is a valid and good solution in contrast to the performance impacts which @commonsensesoftware has described.

commonsensesoftware: @Discjoggy have you tested the side effects using this branch? In looking at the stack trace, it seems related, but perhaps not the same. I would not expect an incomplete type (e.g. **TypeBuilder**) to be the `type` in the line you've highlighted. If that's actually a possibility, then there's a different bug that needs to get addressed.

The **OriginalTypeAttribute** was introduced when I added support for query options, though it potentially has usefulness in other scenarios too. Effectively, there needs to be a way to _backtrack_ from a dynamically generated type back to the original **Type**. This can be used to look it up in the EDM or other purposes.

@LuukN2  you may have noticed that I fixed another related, but previously undetected bug where the attributes declared by the original type were not emitted on the new type. Before that we only copied attributes on properties. Type attributes are may be required for query options and other scenarios.

Discjoggy: > have you tested the side effects using this branch?

I did not notice any side effects yet, all my tests working properly and swagger is working too, but I will spend some more time today.

> If that's actually a possibility, then there's a different bug that needs to get addressed.

I am not really sure about this. I am working with EF Core, Complex Types with no identifiers, nested references, and yep also with bad circular references which gets "value converted" by EF Core, etc.
So my project is very complex and it is hard to quickly look into specific scenarios which I could test, sorry.

> Effectively, there needs to be a way to backtrack from a dynamically generated type back to the original Type

Well, what I've tried is to comment and ignore all error relevant things of the models.
Then I tried to add a complex type with no relation and when I tried to call the SwaggerUI this exception occured:

```txt
System.TypeLoadException: Could not load type '<MODEL>' from assembly 'T0930853bc9ff460da400ea7e1a357a05.DynamicModels, Version=0.0.0.0, Culture=neutral, PublicKeyToken=null'.
   at System.Signature.GetSignature(Void* pCorSig, Int32 cCorSig, RuntimeFieldHandleInternal fieldHandle, IRuntimeMethodInfo methodHandle, RuntimeType declaringType)
   at System.Reflection.RuntimeMethodInfo.FetchNonReturnParameters()
   at System.Reflection.RuntimeMethodInfo.GetParameters()
   at Newtonsoft.Json.Serialization.DefaultContractResolver.GetCallbackMethodsForType(Type type, List`1& onSerializing, List`1& onSerialized, List`1& onDeserializing, List`1& onDeserialized, List`1& onError)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.ResolveCallbackMethods(JsonContract contract, Type t)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.InitializeContract(JsonContract contract)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.CreateObjectContract(Type objectType)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.CreateContract(Type objectType)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
   at Newtonsoft.Json.Serialization.DefaultContractResolver.ResolveContract(Type type)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaRegistry.CreateSchema(Type type, Queue`1 referencedTypes)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaRegistry.CreateArraySchema(JsonArrayContract arrayContract, Queue`1 referencedTypes)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaRegistry.CreateInlineSchema(Type type, Queue`1 referencedTypes)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaRegistry.CreatePropertySchema(JsonProperty jsonProperty, Queue`1 referencedTypes)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaRegistry.CreateObjectSchema(JsonObjectContract jsonContract, Queue`1 referencedTypes)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaRegistry.CreateInlineSchema(Type type, Queue`1 referencedTypes)
   at Swashbuckle.AspNetCore.SwaggerGen.SchemaRegistry.GetOrRegister(Type type)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.CreateResponse(ApiResponseType apiResponseType, ISchemaRegistry schemaRegistry)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.CreateResponses(ApiDescription apiDescription, ISchemaRegistry schemaRegistry)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.CreateOperation(ApiDescription apiDescription, ISchemaRegistry schemaRegistry)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.CreatePathItem(IEnumerable`1 apiDescriptions, ISchemaRegistry schemaRegistry)
   at System.Linq.Enumerable.ToDictionary[TSource,TKey,TElement](IEnumerable`1 source, Func`2 keySelector, Func`2 elementSelector, IEqualityComparer`1 comparer)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.CreatePathItems(IEnumerable`1 apiDescriptions, ISchemaRegistry schemaRegistry)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwagger(String documentName, String host, String basePath, String[] schemes)
   at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.Invoke(HttpContext httpContext, ISwaggerProvider swaggerProvider)   at Microsoft.AspNetCore.Builder.Extensions.MapWhenMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Builder.RouterMiddleware.Invoke(HttpContext httpContext)
   at Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware.Invoke(HttpContext context)
```

Hint: Every step above is working in `Beta1`. 
So since `Beta2` I was not able to get OData work together with Swagger, but yes of course this can be a seperate issue because this is some days ago and maybe there are some other reasons, I will spend some more time and open seperate issues if @LuukN2 disagrees.

LuukN2: @Discjoggy 
Actually, I've been looking for this exception online and it seems there's also 
[a stackoverflow post with a user having the same exception.](https://stackoverflow.com/questions/53705270/swagger-crashes-with-circular-model-references/53738397) They say it reproduces when there are circular references so this only occurs when substitution has taken place. 

I can't get this exception to reproduce on my branch but another dev did get this exception when checking out my branch a few days ago. Specifically, a type could not be loaded from the generated assembly just like the exception you posted.

If you check list of types in the generated assembly before the exception is thrown is it just this type that is missing?

Another dev checking out my branch does not get this exception.
It may be that there are some DLLs still being referenced.

commonsensesoftware: @Discjoggy many of these improvements by @LuukN2 didn't merge until Beta 2+. Do you know if this exception persists when the CLR type correctly matches the EDM type? This should result in no substitution taking place. If that scenario is not broken, then there is a feasible workaround, which was always required before the substitution feature was introduced. If that's broken, then there's a higher severity issue at play because there's a blocking bug.

@LuukN2 and I flushed out a lot of scenarios in the Beta that hadn't previously been covered. We've added a number of unit tests to cover those scenarios, so I'm certainly curious as to the edge cases we're missing. I'm wondering if this is a product of the source types coming from other assemblies, which are not explicitly tested. I would have expected type generation to fail if that was the case.

@Discjoggy since you have a repro are you able to confirm whether the source **Type** that causes the **TypeLoadException** is _complete_? You can check by testing whether the type is **TypeBuilder**, which has the inheritance hierarchy `TypeInfo → Type`. Oddly, the stack trace is indicating that there is an issue reflecting on method information, but no methods are considered in type substitution. @LuukN2 it would appear that there is one or more instances where `TypeBuilder.GetTypeInfo()` is not being called somewhere in the **DefaultModelTypeBuilder**.

As an aside, the NuGet packages now have [SourceLink](https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink) enabled, which should make debugging and troubleshooting easier.

Discjoggy: I will try to keep this answer short and this will probably the last because it would make sense to open seperate issues for that, but you can tell me what to do after you've read this.

@commonsensesoftware
Yesterday I had not enough time to test the matchings in detail. I just quickly looked into the registered models and types of the models and everything was looking right for me.

@LuukN2
Indeed circular references tends to problems very quickly and probably this is a logical mistake by myself which I have to fix. This is why I removed everything to a very basic level to test different scenarios and i came to this is very shrinked abstraction of one of my use cases:

```cs
public class EntityTypeX
{
    public Guid Id { get; set; }
    public ComplexTypeA A { get; set; }
    public ComplexTypeB B { get; set; }
}

public class ComplexTypeA
{
    public float Value { get; set; }
}

public class ComplexTypeB
{
    public string Name { get; set; }
    public ComplexTypeA ComplexValue { get; set; } // throws System.TypeLoadException
}

```

This [Issue](https://stackoverflow.com/questions/50682776/how-to-support-a-nested-open-complex-type-in-the-odata-c-sharp-driver) on stackoverflow looks very similar to this but even with many variations / tries with the `ODataModelBuilder` it was not able to register this relation / types.

Some side effects and information:

- With `Beta2` the `FromOdataRouteAttribute` is no longer required for Actions and Functions
- With `3.1.0` I get a 404 when I try to call the `$metadata` document
- As described before, *nesting of complex types* / *using complex properties in complex types* seems to be forbidden (need to check this in the OData specs)

Discjoggy: Another side effect comes with the use of this merge. The SwaggerDoc works but all the complex properties of the EntityTypes which would throw this exception are missing in the doc.

test/Microsoft.AspNetCore.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
```diff
@@ -355,6 +356,29 @@ public void substitute_should_generate_types_for_actions_with_the_same_name_in_d
...
+            var addressType = typeof(Address);
+
+            var context = NewContext( modelBuilder.GetEdmModel() );
+
+            // act
+            addressType.SubstituteIfNecessary( context );
+            var substitutionType = originalType.SubstituteIfNecessary( context );
+
+            // assert
+            Assert.False( substitutionType is TypeBuilder );
```
commonsensesoftware: This isn't wrong, but for consistency let's continue to use Fluent Assertions, which would look like this:

```c#
substitutionType.Should().NotBeOfType( typeof( TypeBuilder ) );
```

>Don't forgot to format the document ;)

test/Microsoft.AspNet.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
```diff
@@ -377,6 +378,29 @@ public void substitute_should_get_attributes_from_property_that_has_attributes_t
...
+            var addressType = typeof(Address);
+
+            var context = NewContext( modelBuilder.GetEdmModel() );
+
+            // act
+            addressType.SubstituteIfNecessary( context );
+            var substitutionType = originalType.SubstituteIfNecessary( context );
+
+            // assert
+            Assert.False( substitutionType is TypeBuilder );
```
commonsensesoftware: Same as below

src/Microsoft.AspNet.OData.Versioning.ApiExplorer/ReleaseNotes.txt
```diff
@@ -1 +1,2 @@
-﻿Fixed query options for non-action POST
\ No newline at end of file
+﻿Fixed query options for non-action POST
```
commonsensesoftware: Apologies, I probably wasn't clear on this. You need only replace the release notes. Since the list can get very long, the package release notes are whatever was patched in the published package. The next release on GitHub will have the rollup of everything. I moved to using a text file so it would be easier to diff the file for changes over time. You can simply SHIFT+DEL the first line. Thanks.

commonsensesoftware:
@LuukN2 in looking at where there might be an incomplete type, it appears there is a real chance this can happen in `DefaultModelTypeBuilder.ResolveDependencies`. If I'm not mistaken, there is no scenario where this method should return a **TypeBuilder**, but the current implementation does allow for it. I _think_ we have a repro for it now. Another user has also reported the issue in #431.

It would also appear that once you reach the bottom of the graph, you need the type for building dependencies back up. I believe that's the purpose of `unfinishedTypes`. I _think_ this means that in addition to not returning **TypeBuilder** from `ResolveDependencies`, the type _may_ need to be updated with the completed type. Something like following:

```c#
unfinishedTypes.AddOrUpdate( typeKey, typeBuilder, ( key, tb ) => tb.CreateTypeInfo() );
```

and probably near the end of the method. Should this be the case, there would be a little more refactoring since the value in the dictionary is currently **TypeBuilder** instead of **Type** or **TypeInfo**.

I do think there are two distinct issues here. If you want to tie this PR off and start another for the second issue, I'm fine with that. The next package release should ultimately include both fixes though. Let me know how you'd like to proceed (so I can merge this). Thanks.

LuukN2:
Right now **ResolveDependencies** can still return a **TypeBuilder** because future dependencies may still have not been passed.

One way to refactor this is to only call **ResolveDependencies** once all needed structured types have been passed. This will also make it a lot easier to debug later because you will be able to check if the list of property dependencies is not empty.

This will probably make it run faster and simplify the code because it would mean there be no need to iterate the dependencies each time a structured type needs to be substituted. The **ResolveDependencies** method could just iterate past each item in the list of property dependencies and afterwards build every type. The **generatedTypes** and **unfinishedTypes** can then also become one Dictionary for resolving the properties.

Right now I'll fix the release notes and reformat the unit tests and commit that.

On a side note:
It seems spaces aren't added between keywords and parentheses since updating Rider when formatting. :sob:

test/Microsoft.AspNetCore.OData.Versioning.ApiExplorer.Tests/AspNet.OData/DefaultModelTypeBuilderTest.cs
LuukN2: The problem with this is that it checks for the type that the **TypeBuilder** is building, meaning that the test will succeed even if it is a **TypeBuilder**.

I've tried:
`NotBeOfType<TypeBuilder>();`
`NotBe<TypeBuilder>();`
`NotBeDerrivedFrom<TypeBuilder>();`

With each of these methods the test will succeed even if the type is a **TypeBuilder**.
Maybe there is a method or property that extracts **TypeBuilder** from the type or that indicates it is a **TypeBuilder** that could be used instead, but unless there is a Fluent assertion that works for this I think it's best to leave it as is.

commonsensesoftware: Interesting... could be a bug in Fluent Assertions. The `is` keyword is basically the same thing as `Type.IsAssignableFrom`. Correctness is more important here. I'll consider revisiting this some time in the future the next time I update Fluent Assertions. Thanks.

commonsensesoftware:
@LuukN2 just confirming. You're ready to close this PR out and a separate PR will spin up to investigate/improve **ResolveDependencies** (or whatever is causing incomplete types)? I looks like things are good to go.

LuukN2:
Yeah, I think this one is ready to be closed and I'll make a seperate one for refactoring and looking into the issue people are having later.

commonsensesoftware:
Perfect! Thanks.

BorisWilhelms:
@commonsensesoftware: Is there any chance that this fix can/will be backported to Microsoft.AspNetCore.OData.Versioning.ApiExplorer v3.0, which is the last version that is compatible with AspNetCore 2.1 (LTS)?

commonsensesoftware:
It's possible, but for an army of one like me, it is _painful_. I won't say _"no_", but it's likely to be lower in my priority list. I've got quite a few open investigations at the moment.

Out of curiosity, have you tried moving to 2.2 while using 2.1 compatibility and _legacy_ routing? The last couple of people that have asked for some type of backport were able to move forward by doing this. That _should_ be something you can verify pretty quickly. If that works for you, it's likely a much better and faster alternative than waiting for me to catch up.

If it doesn't work, then I can queue up a backport. I knew this _could_ happen, which is why I spit 3.0 and 3.1 apart even though they were released back-to-back and are nearly functionality equivalent.

BorisWilhelms:
Thank you for your quick answer and all your work! It is more an internal process/compliance and timeline topic that we are staying on LTS than a technical topic. I understand that this has low prio for you and I will discuss this topic internally.

Issue #431, opened 2019-01-04
The invoked member is not supported before the type is created.


albertwoo:
Step to reproduce:
1. In /src/samples/aspnetcore/SwaggerODataSample/Models/Person.cs, add property `public Bug Bug { get; set; }` to Person class. Bug class is like:
```c#
    public class Bug
    {
        public int Id { get; set; }
        public string Detail { get; set; }
    }
```
2. Run the project and I got Exception1.
3. This line https://github.com/Microsoft/aspnet-api-versioning/blob/509bdfd63c4a9a7fc54c6151a405f07bd9adceb3/src/Common.OData.ApiExplorer/AspNet.OData/StructuredTypeResolver.cs#L20 has a bug. we should change to something like:
```c#
var structuredType = structuredTypes.FirstOrDefault( t => type.FullName == t.GetClrType( model )?.FullName );
```
4. After we fix this, the project can run now, but in the swagger ui, if we switch to v3, we got Exception2

Exception1:
```
   at System.Reflection.Emit.TypeBuilder.GetCustomAttributes(Type attributeType, Boolean inherit)
   at System.Attribute.GetCustomAttributes(MemberInfo element, Type type, Boolean inherit)
   at System.Attribute.GetCustomAttribute(MemberInfo element, Type attributeType, Boolean inherit)
   at System.Reflection.CustomAttributeExtensions.GetCustomAttribute[T](MemberInfo element, Boolean inherit)
   at Microsoft.AspNet.OData.StructuredTypeResolver.GetStructuredType(Type type) in C:\Users\binwwu\Documents\Github\aspnet-api-versioning\src\Common.OData.ApiExplorer\AspNet.OData\StructuredTypeResolver.cs:line 28
   at Microsoft.AspNet.OData.Builder.ODataAttributeVisitor.Visit(ApiDescription apiDescription) in C:\Users\binwwu\Documents\Github\aspnet-api-versioning\src\Common.OData.ApiExplorer\AspNet.OData\Builder\ODataAttributeVisitor.cs:line 63
   at Microsoft.AspNet.OData.Builder.ODataValidationSettingsConvention.ApplyTo(ApiDescription apiDescription) in C:\Users\binwwu\Documents\Github\aspnet-api-versioning\src\Microsoft.AspNetCore.OData.Versioning.ApiExplorer\AspNet.OData\Builder\ODataValidationSettingsConvention.cs:line 38
[...]

PR #433, opened 2019-01-09
Refactored the DefaultModelTypeBuilder

LuukN2:
Changes:
- Now resolves all dependencies on the first call to `NewStructuredType`.
- All types will always be built, see `ResolveDependencies`.
- Fixed a bug where a collection dependency would add itself as the type rather than the dependency.

I refactored the **DefaultModelTypeBuilder** to resolve all dependencies on the first call.
This is done by iterating through all structured types in the EdmModel. In most cases this will be faster assuming most entity types are used in the api endpoints. This prevents iterating through all generated types to resolve again for each structured type.
Because all structured types have been passed and there are no later dependencies all types of **TypeBuilder** can safely be built.

Hopefully this resolves the exception people are having. If not, we can at least exclude remaining types of **TypeBuilder** to be causing the issue.

Note: The **DefaultModelTypeBuilder** no longer requires passing a clr type along with the structured type but I left the **IModelTypeBuilder** unchanged since it's a public interface.

### DIFF
--- a/src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/DefaultModelTypeBuilder.cs
@@ -28,10 +28,9 @@
     {
         static readonly Type IEnumerableOfT = typeof( IEnumerable<> );
         readonly ConcurrentDictionary<ApiVersion, ModuleBuilder> modules = new ConcurrentDictionary<ApiVersion, ModuleBuilder>();
-        readonly ConcurrentDictionary<EdmTypeKey, TypeInfo> generatedEdmTypes = new ConcurrentDictionary<EdmTypeKey, TypeInfo>();
-        readonly ConcurrentDictionary<EdmTypeKey, TypeBuilder> unfinishedTypes = new ConcurrentDictionary<EdmTypeKey, TypeBuilder>();
+        readonly ConcurrentDictionary<ApiVersion, ConcurrentDictionary<EdmTypeKey, TypeInfo>> generatedEdmTypesPerVersion = new ConcurrentDictionary<ApiVersion, ConcurrentDictionary<EdmTypeKey, TypeInfo>>();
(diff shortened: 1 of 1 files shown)


commonsensesoftware:
This is looking great. Less **is** more! ;)

It definitely seems the **clrType** parameter is no longer necessary, but the initial lookup doesn't _hurt_. I'll consider removing it in the next major version. Although they _can_, I doubt anyone will ever venture down this path to create their own model builder. The future impact of removing this parameter in a major release is probably really low.

The only _real_ feedback I have is whether you were able to create a repro based on the information provided in the previous PR and bug #431? If you weren't able to, I won't hold up this PR because I _think_ this is highly likely the culprit. It would be _nice_ to have a test that confirmed the issue and demonstrates that this change fixed it. It'd also be a nice defensive against any future regression.

If you're not able to recreate the conditions, perhaps @Discjoggy or @albertwoo can help verify. Worst case, I'll put a beta release out for folks to try before committing to an official patch asserting that this change fixes everyone.

Thanks for the assist on this! :D

Discjoggy:
To keep this answer short
- No exceptions
- All tests shine green
- Swagger is working
- All models are looking good except `ICollection<T>` Properties which are not shown in Swagger and I would like to verify this quickly, but..
- The `$metadata` document is still missing (This is probably a different issue and I will create it today)

commonsensesoftware:
@LuukN2 ... ah ... misunderstood that #431 was fixed in the previous PR. I agree with your assessment. It certainly appeared that **TypeBuilder** was indeed being used instead of **Type**. The only remaining place I could find where that _might_ happen was within `ResolveDependencies`, which you have now refactored. I think we're good to from that perspective.

@Discjoggy, you'll need to provide a scenario where `ICollection<T>` doesn't work. Any type that is covariant with `IEnumerable<T>` should _just work_. As far as the `$metadata` endpoint is concerned, that seems to be unrelated and am I also unable to repro the scenario. I see you opened #435 to track it separately. We'll hash out the details there.

Issue #441, opened 2019-01-14
deprecated property on operations in swagger.json


alberto-riggi:

alberto-riggi:
Hi,
I saw there was already a similar issue opened regarding the usage of deprecated on actions, but I think my question is different. 
I was expecting that marking a service version as deprecated would have caused to also add the deprecated property on each operation in the swagger.json description (the same behavior as using the obsolete attribute on an action).

I'm kind of new to versioning and swagger. 
Is it intentional? Am I missing something?

Regards

commonsensesoftware:
There is no **direct** relationship or dependencies between API versioning and Swagger. The provided API Explorer can tell you whether a specific action is deprecated for a particular API version or whether the entire API version is deprecated (e.g. for a given API version, all containing actions are deprecated).

If you want a specific operation in Swagger to be marked as deprecated, it would be your responsibly to indicate that in your Swagger generator of choice (ex: Swashbuckle, NSwag, etc). API Versioning provides you with the necessary information to determine whether said API is deprecated, but it's up to you to take advantage of that information. The API Explorer can be used for other scenarios besides Swagger generation. The exact method in which you tie the two together depends on which Swagger generator you are using.

I hope that helps clear things up.

commonsensesoftware:
Correct. This is because:

1. API Explorer doesn't know anything about Swagger
2. Swagger generators (like Swashbuckle) **only** know about information provided by the default API Explorer interfaces
   a. `ApiVersion` and `IsDeprecated` are not standard properties on `ApiDescription`

While it's totally convenient for the two to marry up, it doesn't really make sense for either side to take a hard dependency on the other. The most ideal situation would be for someone in the community (wink, wink) to create a new package that did this pairing.

Although I don't plan to support directly wiring up Swagger information, I have absolutely no problems providing samples as to _how_ it can/should be done. There was no existing reason that couldn't have been done today, I just hadn't done it (and I'm not exactly a Swagger expert). You didn't indicate which platform you're using, but I've updated all of the Swagger examples to set the deprecated member of the Swagger operation.

[...]

PR #454, opened 2019-01-29
Add an IsDeprecated Extension Method

commonsensesoftware:
Add an IsDeprecated extension method, which will simplify API Explorer use such as in Swagger document generation. Related to #441.


esurmay:
what about deprecate operations and not all API service?

commonsensesoftware:
I'm not sure I follow. Are you asking if this extension method would work on say - a UI controller that happened to have some APIs on it? In this case, it _should_ work on ASP.NET Core, though I've never tried or seen anyone with such a setup. On Web API it would not work because all controllers are service-only and using UI controllers (e.g. MVC on the legacy stack) is not supported.

This extension method would meant to be purely be a shortcut to something you could already do, but might not know how to.

PR #478, opened 2019-03-21
fallback to a sane default when edm parameters and clr parameters disagree

hippiehunter:
I have a custom odata routing convention to deal with optional parameters, unfortunately this leaves me in a position where the clr parameters disagree with the declared edm parameters. The routing convention will dispatch and translate where appropriate but attempting to build the swagger doc fails as a result. I have an additional PR that i will be sending to the odata\odata.net repo, The odata path templates produced in this workflow end up having a slightly different shape that just needed to be handled.

### DIFF
--- a/src/Common.OData.ApiExplorer/AspNet.OData/Routing/ODataRouteBuilder.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/Routing/ODataRouteBuilder.cs
@@ -222,12 +222,14 @@
                 var actionParameters = Context.ParameterDescriptions.ToDictionary( p => p.Name, StringComparer.OrdinalIgnoreCase );
                 var parameter = parameters.Current;
                 var name = parameter.Name;
+                actionParameters.TryGetValue( name, out var mappedParameter );
 #if WEBAPI
-                var routeParameterName = actionParameters[name].ParameterDescriptor.ParameterName;
+                var routeParameterName = mappedParameter?.ParameterDescriptor?.ParameterName ?? name;
 #elif API_EXPLORER
-                var routeParameterName = actionParameters[name].ParameterDescriptor.Name;
+
+                var routeParameterName = mappedParameter?.ParameterDescriptor?.Name ?? name;
 #else
-                var routeParameterName = actionParameters[name].Name;
+                var routeParameterName = mappedParameter?.Name ?? name;
 #endif
 
                 builder.Append( '(' );
@@ -239,12 +241,14 @@
                 {
                     parameter = parameters.Current;
                     name = parameter.Name;
+                    actionParameters.TryGetValue( name, out mappedParameter );
 #if WEBAPI
-                    routeParameterName = actionParameters[name].ParameterDescriptor.ParameterName;
+                    routeParameterName = mappedParameter?.ParameterDescriptor?.ParameterName ?? name;
 #elif API_EXPLORER
-                    routeParameterName = actionParameters[name].ParameterDescriptor.Name;
+
+                    routeParameterName = mappedParameter?.ParameterDescriptor?.Name ?? name;
 #else
-                    routeParameterName = actionParameters[name].Name;
+                    routeParameterName = mappedParameter?.Name ?? name;
 #endif
                     builder.Append( ',' );
                     builder.Append( name );


hippiehunter:
@commonsensesoftware Once #502 is resolved can you take a look at this?

commonsensesoftware:
Sure thing. I've been really backed up lately.

commonsensesoftware:
Apologies for the long delay. I'm finally circling back around to this.

I'm not sure I fully understand the scenario that this is addressing. What exactly do you mean by:

>_"...the clr parameters disagree with the declared edm parameters"_

Are you referring to action parameters that are not part of the EDM? Those should be added and processed as query string parameters, unless I'm missing something.

I'm trying to get a better understanding of the situation to make sure we're going down the right path. It appears that the parameter names are not matching up. All action parameters _should_ be a superset of any parameters defined by the EDM action or function.

OData has some advanced edge cases so I won't be all that surprised if I missed something. Thanks.

hippiehunter:
So one of the tasks that my calling convention does, is translation of some odata query operators into parameters, This ends up meaning that the parameters declared to the EDM are not the same as the ones declared to the CLR. Specifically I end up passing the parameters into a single attributed argument that contains all of the things declared to the EDM. So its not a case of having fewer EDM arguments, its that I have a bunch of EDM arguments and no matching CLR arguments.

commonsensesoftware:
Sounds kind of like custom _model binding_. Since there aren't any new tests that illustrate your scenario, can you provide the world's simplest example here? I'm trying to wrap my _small squirrel brain_ around exactly what this looks like.

Support for documenting query options is relatively new. I guess I never really considered a scenario as wild as translating the parameters to mean something else, but I suppose that's valid. There may need to be a configuration option to not document query options in this type of setup.

hippiehunter:
I have some docs on this mechanism here https://github.com/Synergex/HarmonyCore/wiki/Adapters does this provide enough of an example?

commonsensesoftware:
That definitely provides some context - interesting. It would appear that this framework calls `ODataQueryOptions.ApplyTo` somewhere in the pipeline, then runs it through a custom **ExpressionTreeVisitor**, and decomposes the constituent parts into a _model-bound_ object defined by a user mapped via some custom attributes.

Let me digest all of this and see if I don't have any more questions.

src/Common.OData.ApiExplorer/AspNet.OData/Routing/ODataRouteBuilder.cs
```diff
@@ -222,12 +222,14 @@ void AppendParametersFromConvention( StringBuilder builder, IEdmOperation operat
                 var actionParameters = Context.ParameterDescriptions.ToDictionary( p => p.Name, StringComparer.OrdinalIgnoreCase );
                 var parameter = parameters.Current;
                 var name = parameter.Name;
+                actionParameters.TryGetValue( name, out var mappedParameter );
```
commonsensesoftware: I'm still not sure I 100% grock the full use case, but I don't see anything glaring where this change is specific to your scenario. It appears to be a rational fallback, even if it shouldn't happen.

Let's simply change the entire setup to:

```c#
                var routeParameterName = name;
                
                if ( actionParameters.TryGetValue( name, out var mappedParameter ) )
                {
#if WEBAPI
                    routeParameterName = mappedParameter.ParameterDescriptor.ParameterName;
#elif API_EXPLORER
                    routeParameterName = mappedParameter.ParameterDescriptor.Name;
#else
                    routeParameterName = mappedParameter.Name;
#endif
                }
```

src/Common.OData.ApiExplorer/AspNet.OData/Routing/ODataRouteBuilder.cs
```diff
@@ -239,12 +241,14 @@ void AppendParametersFromConvention( StringBuilder builder, IEdmOperation operat
                 {
                     parameter = parameters.Current;
                     name = parameter.Name;
+                    actionParameters.TryGetValue( name, out mappedParameter );
```
commonsensesoftware: The change down here should obviously be the same.

commonsensesoftware:
Seems like you might be busy. I've incorporated the same changes in #521 so I'm going to close this PR out. I'm about to start the release build so I can publish the packages that will include this change. Thanks.

hippiehunter:
thanks, sorry for the slowness.

commonsensesoftware:
No problem. It was small change so I just lifted it over for you. The latest packages with this change have been published. Cheers!

PR #512, opened 2019-07-08
Support for properties renamed with [DataMember] in OData models

brainboost:
Allows OData model properties to be renamed with DataMemeber attribute

PR #520, opened 2019-07-17
Dependency properties should keep custom attributes

brainboost:
I found that dependencies were created without custom attributes that led to lost properties in DataContract as only those marked with [DataMember] were shown. This addition fixes it. Also, I provide unit tests for attributed dependencies.


commonsensesoftware:
Good find. Looks like you've covered the bases. Does this PR supplant #512? I haven't merged it yet, but I was getting ready to.

brainboost:
Yes, it does. This one includes all changes as I created that branch from the feature branch of #512

PR #543, opened 2019-09-08
Add properties reordering feature to the swagger generated OData models

brainboost:
This PR adds example of using Swashbuckle schema filter applied to the OData Asp.Net Core versioning sample in order to allow reordering of model properties using the `DataMemberAttribute`.
Usage: set the Order in [DataMember] with non-negative number to apply property sort order
```
    [DataContract]
    public class Address
    {
        /// <summary>
        /// Gets or sets the street address.
        /// </summary>
        /// <value>The street address.</value>
        [DataMember( Order = 0 )]
        public string Street { get; set; }

        /// <summary>
        /// Gets or sets the address city.
        /// </summary>
        /// <value>The address city.</value>
        [DataMember]
        public string City { get; set; }

        /// <summary>
        /// Gets or sets the address zip code.
        /// </summary>
        /// <value>The address zip code.</value>
        [DataMember( Name = "zip", Order = 1 )]
        public string ZipCode { get; set; }
    }
```
In the example above, property order in swagger doc will look like this:
**Address**
- street | _string_  Gets or sets the street address.
- zip | _string_  Gets or sets the address zip code.
- city | _string_  Gets or sets the address city.

> NOTE: the ZipCode property has been renamed as well as reordered. Non-reordered properties get initially defined order from source.

### DIFF
--- a/samples/aspnetcore/SwaggerODataSample/Models/Address.cs
+++ b/samples/aspnetcore/SwaggerODataSample/Models/Address.cs
@@ -18,7 +18,7 @@
         /// Gets or sets the street address.
         /// </summary>
         /// <value>The street address.</value>
-        [DataMember]
+        [DataMember( Order = 0 )]
         public string Street { get; set; }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


commonsensesoftware:
I'm trying to wrap my brain around how this is beneficial to others. The ordering of data members seems highly subjective; especially for an example. I intentionally try to keep things about specific Swagger implementations to a minimum. This change doesn't really appear to provide additional information specific to versioning an API.

It's also worth mentioning that I keep the examples as congruent between the platforms as possible so it will ease the transition from Web API to Core. Suggestions and changes to examples need to be the same in both.

This kind of feels like it would be better served in one of the Swagger repos or Gist, no?

brainboost:
I agree with @commonsensesoftware, it is rather specific feature, bound to the odata and swagger in implementation. My intention was to share it with people who may concern, as I already spent time on it. I absolutely won't mind if you dismiss this PR. Thanks for your time and advice, 
Regards

PR #544, opened 2019-09-11
Update Package Icon Url

Varorbc:
### It is recommended to use the .net icon instead of the Microsoft icon

image

https://go.microsoft.com/fwlink/?LinkID=288859

http://go.microsoft.com/fwlink/?LinkID=288890

### DIFF
--- a/build/nuget.props
+++ b/build/nuget.props
@@ -4,7 +4,7 @@
  <PropertyGroup Label="NuGet">
   <RepositoryType>git</RepositoryType>
   <RepositoryUrl>https://github.com/Microsoft/aspnet-api-versioning</RepositoryUrl>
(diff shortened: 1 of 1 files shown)


commonsensesoftware:
I need to follow up on this for compliance. That is the icon for the **[.NET Foundation](https://dotnetfoundation.org/
)**, which this project is not a part of - yet. Many of the large product teams are now part of the .NET Foundation, but since I'm running this myself, I haven't gone through all the steps to make that happen.

There are compliance people on NuGet that periodically check for this type of stuff. Changing the icon isn't a huge deal, but it can't be misrepresentation. Naturally, if this project becomes part of the .NET Foundation, the icon will be expected to change.

PR #548, opened 2019-09-21
Sample .NETCORE RC1

ivanpaulovich:
Hi @commonsensesoftware,

First of all, congratulations for the awesome project!

I added a sample with `.NET Core 3.0 RC1` and `Swashbuckle 5.0.0-rc2`.
I think it will help the ones using the latest packages.

Let me know if I need to change something else.

Regards,
Ivan


commonsensesoftware:
Thanks for the contribution. I'm not sure, but I suspect you may not have realized that there are already .NET Core 3.0 examples in the [3.0.1 Release](../tree/release/3.0.1/samples/aspnetcore) branch. This will be updated to RTM in the near future and will become the default navigation once it's merged to **master**.

If there is something else separate or worth keeping, I'm happy to consider it.

emilssonn:
Might be the wrong place to comment this but. The example in this PR uses `IConfigureOptions` with dependency injection. This removes the warning about `services.BuildServiceProvider()`. But this warning might not be a problem since it is in a using statement?

ivanpaulovich:
Hi @commonsensesoftware!

My bad I miss that branch! I will check there if there is something I can contribute. Thank you for the great work!

Issue #542, opened 2019-09-04
SwaggerODataSample error


spaasis:
Hi, and thanks for the brilliant library!

I was studying the SwaggerODataSample and noticed that doing a GET to `/api/Orders?api-version=2.0&%24count=false` (default values from swagger) returns an error:

> The query specified in the URI is not valid. The entity set 'Orders' cannot be used for $count.

The query works fine if we remove `Count` from the `AllowedQueryOptions`.

I don't yet know enough about OData to propose a fix, so just a heads up this time.

spaasis:
On a sidenote, when I navigate to V2 on swagger and click either of the `Order`-methods, both of them open/close. The same happens with `People`. Does anyone else see this behavior?

commonsensesoftware:
Apologies for the long delay in the response.

First, you likely need to enable the global `$count` operation in the setup via `routeBuilder.Count()`. It's not clear why it doesn't work when it's enabled at the action level. I suspect it's a bug in OData.

The Swagger examples use Swashbuckle. Oddities in the UI behavior would need to be directed in their repo. I'll be updating the versions shortly and I suspect that issue may have been fixed.

I hope that helps. Thanks.

spaasis:
Added a PR https://github.com/microsoft/aspnet-api-versioning/pull/567  to fix the bug, and can confirm that the UI stuff is fixed in at least some later version of Swashbuckle. Closing now.

PR #567, opened 2019-10-28
fix: #542 in SwaggerODataSample

spaasis:
$count was not enabled globally, so using it in queries caused a 400 error


commonsensesoftware:
Thanks for spending the time on this. Honestly, I feel like this is a bug in OData. What's the purpose of using `[EnableQuery]` if it only works when you enable to the global query option? I can't really say, but it's worth adding this so it works.

Ultimately, this setting needs to be enabled in both the ASP.NET Core and Web API examples. I've taken the liberty of updating both. I just finishing merging the 4.0 bits so you should see it available now.

I'm going to close this PR as it's no longer necessary, but I referenced in my changes. Thanks.

spaasis:
Oh right, didn't even think about the webapi side of things :)
Yeah, the config is quite confusing in OData. It's not too well documented either...
I opened this to see if it could be made a bit easier: https://github.com/OData/WebApi/issues/1942 

Cheers for the 4.0 release!

PR #632, opened 2020-05-13
correct ApiVersionDescription.IsDeprecated docs

lancehilliard:
When some API versions are partially deprecated, IsDeprecated returns false (not true).


commonsensesoftware:
Awesome thanks

PR #642, opened 2020-06-22
Fixed typos

ronwarner:
While looking through examples, I took the time to fix some typos.


ronwarner:
Without visibility to the ADO build pipeline, I cannot tell why if failed.

commonsensesoftware:
It's not you, it's me. There's something up with the pipeline, but I'm not sure what. If everything built and all tests passed locally, you can be almost assured that things are fine on the server end. The pipeline is **super** vanilla; especially for PRs. The main reason it's there is for the internal MS code signing.

I've been really overloaded the past few months, but I'm planning to get back into all of the work here ASAP. Thanks for fixing my typos. ;)

commonsensesoftware:
Finally figured out the issues. It was a cascading set of problems related to SourceLink and code analysis. Looks like I'm back in business. Thanks.

PR #668, opened 2020-10-01
Improve ApiVersioning.sln

SaurabhDev338:


commonsensesoftware:
The solution file regularly changes the version number over time as more recent versions of Visual Studio touch it. As far as I can tell, this change now aligns to what's committed. While I appreciate you taking the time to make this change, it seems it's no longer necessary. The latest version of the repository should now include this change. Thanks.

Issue #705, opened 2020-12-03
System.InvalidOperationException: ValueFactory attempted to access the Value property of this instance. Microsoft.AspNet.WebApi.Versioning


abdkayali:
I'm working on an OData Solution based on ASP.NET, and I came to the point to add nuget packages for api versioning, so I ended up adding the following:

`Microsoft.AspNet.OData.Versioning` v5.0.0
`Microsoft.AspNet.WebApi.Versioning` v4.1.0

Upon building and testing the code, I get the following exception for whatever request:

`Exception Details: System.InvalidOperationException: ValueFactory attempted to access the Value property of this instance.`

Stack Trace:
```
[InvalidOperationException: ValueFactory attempted to access the Value property of this instance.]
   System.Lazy`1.CreateValue() +429
   System.Lazy`1.LazyInitValue() +158
   System.Lazy`1.get_Value() +79
   Microsoft.Web.Http.Dispatcher.ApiVersionControllerSelector.GetControllerMapping() in /_/src/Microsoft.AspNet.WebApi.Versioning/Dispatcher/ApiVersionControllerSelector.cs:49
   System.Web.Http.Routing.AttributeRoutingMapper.AddRouteEntries(SubRouteCollection collector, HttpConfiguration configuration, IInlineConstraintResolver constraintResolver, IDirectRouteProvider directRouteProvider) +48
   System.Web.Http.Routing.<>c__DisplayClass1_1.<MapAttributeRoutes>b__1() +57
   System.Web.Http.Routing.RouteCollectionRoute.EnsureInitialized(Func`1 initializer) +61
   System.Web.Http.Routing.<>c__DisplayClass1_0.<MapAttributeRoutes>b__0(HttpConfiguration config) +91
   System.Web.Http.HttpConfiguration.ApplyControllerSettings(HttpControllerSettings settings, HttpConfiguration configuration) +70
   System.Web.Http.Controllers.HttpControllerDescriptor.InvokeAttributesOnControllerType(HttpControllerDescriptor controllerDescriptor, Type type) +165
   System.Web.Http.Controllers.HttpControllerDescriptor.InvokeAttributesOnControllerType(HttpControllerDescriptor controllerDescriptor, Type type) +54
   System.Web.Http.Controllers.HttpControllerDescriptor..ctor(HttpConfiguration configuration, String controllerName, Type controllerType) +105
[...]

PR #706, opened 2020-12-05
Fix a bug in ApiVersionControllerSelector.cs

abdkayali:
This patch should prevent the following exception that might occur in unspecified conditions:
System.InvalidOperationException: ValueFactory attempted to access the Value property of this instance.

### DIFF
--- a/src/Microsoft.AspNet.WebApi.Versioning/Dispatcher/ApiVersionControllerSelector.cs
+++ b/src/Microsoft.AspNet.WebApi.Versioning/Dispatcher/ApiVersionControllerSelector.cs
@@ -12,6 +12,7 @@
     using System.Web.Http;
     using System.Web.Http.Controllers;
     using System.Web.Http.Dispatcher;
+    using System.Threading;
     using static Microsoft.Web.Http.Versioning.ErrorCodes;
     using static System.StringComparer;
 
@@ -36,7 +37,7 @@
         {
             this.configuration = configuration;
             this.options = options;
-            controllerInfoCache = new Lazy<ConcurrentDictionary<string, HttpControllerDescriptorGroup>>( InitializeControllerInfoCache );
+            controllerInfoCache = new Lazy<ConcurrentDictionary<string, HttpControllerDescriptorGroup>>( InitializeControllerInfoCache, LazyThreadSafetyMode.PublicationOnly );
             controllerTypeCache = new HttpControllerTypeCache( this.configuration );
         }
 


commonsensesoftware:
Thanks for putting forth the effort for a fix. Before I accept, I'll like to understand _why_ this is happening. Do you have repro that you can share? Based on the call stack, I can see that _something_ is causing the cache to be recursively created. I haven't had the chance to analyze the stack further yet. The [original ASP.NET implementation](https://github.com/aspnet/AspNetWebStack/blob/master/src/System.Web.Http/Dispatcher/DefaultHttpControllerSelector.cs#L44) did not use this flag. Ultimately, the cache should not be initialized recursively. That fact that this is happening suggests there's more to this.

abdkayali:
I'll prepare a repro and upload it here when it's ready.
Also I'd like to link the issue #705 where I wrote the stack trace for later reference
Thanks,

abdkayali:
Here's the repro
[Web_Issue705.zip](https://github.com/microsoft/aspnet-api-versioning/files/5652931/Web_Issue705.zip)

abdkayali:
@commonsensesoftware Would you please have a look at the repro? I'm forced to use the clone of solution to get my bug fixed locally. Thanks

commonsensesoftware:
Commenting here because I realized that I closed #705 and missed that there was this PR associated with it.

Your repro project had a few fundamental issues:

1. Don't use `HttpConfiguration.MapAttributeRoutes()`
    1. You don't have any non-OData controllers so it's not necessary
    2. OData **only** uses convention-based routing; mixing with attribute-routing will almost certainly cause issues
2. **CustomizedEnableQueryAttribute** assumes the property `"Number"`, which doesn't exist in any models

If you intend to use the _Help Pages_ feature, then I recommend you add the **Microsoft.AspNet.OData.Versioning.ApiExplorer** package.

Here is your repro project echoed back with the required fixes and comments. I was able to make this work without any library changes.

[Web_Issue705-Working.zip](https://github.com/microsoft/aspnet-api-versioning/files/6736909/Web_Issue705-Working.zip)

If you're encountering other issues, we can dig into it further.

commonsensesoftware:
Thank you for taking the time to submit a PR. The issue turns out to be related to your setup and configuration. I provided a working example without any changes. Since there has been no other follow up since then, I'm to close this PR out. If you feel this is an error, please open a new issue and reference this PR so we can discuss it more first. Thanks.

Issue #746, opened 2021-06-03
Prefer explicit over implicit version in MediaTypeApiVersionReader


DerNamenlose:
When having a `POST` request with `Content-Type` and `Accept` headers, the `MediaTypeApiVersionReader` will always prefer the `Content-Type` header, _even if_ it does not contain any version parameter at all. I'd like to propose to always prefer explicit versions over implicit ones, i.e. if the `Accept` header contains a version parameter and the `Content-Type` does not, then the `Accept` header should be preferred.

We have this use case in our product, where the `Content-Type` of the `POST` request body will not change between different versions (it is a rather stable filter object), but the `Accept` header will, because the response objects differ. The current behavior requires us to introduce a new `Content-Type` for the `POST`, even though its actual contents did not change at all.

I would have a pull request with the changed behavior and appropriate tests ready, if there is interest in that.

commonsensesoftware:
There's been a few debates about this before. I've contemplated changing it a few times, but my better judgement keeps coming back. The version _should_ not be thought of as a separate thing; it _is_ part of the media type. Honestly, it's really part of the content negotiation regardless of whether you version by media type. You have elected to version by media type. Congrats on going the RESTful way! 😉 

These headers have very specific meanings:

- `Content-Type` - What I, the client, am giving you server
- `Accept` - What I, the client, want the server to give me back

While it might seem pedantic, the differences are important. These two headers do not have to match - neither in API version nor media type. This is a perfectly valid request:

```http
POST /order HTTP/2
Host: localhost
Accept: text/csv; q=0.6, application/xml; q=0.4, text/plain
Content-Type: application/json
Content-Length: 42

{"name": "John Doe", "products": [42]}
```

Where I'm going with this is that it is more than reasonable to have `Content-Type: application/json; v=1.0` and `Accept: application/json; v=2.0`. This behavior is not supported out-of-the-box, but it's possible to make it work. You would likely need a custom `InputFormatter` and `OutputFormatter` for each API version (or something like that).

Ultimately, the main challenge here is that there is not a universal way to accommodate all APIs that version this way. Changing formatters requires quite a bit of remapping. This is the same reason that the `application/vnd.json+v1` style isn't intrinsically supported. The format to parse it is not standard and you'd have to remap all the formatters.

[...]

PR #747, opened 2021-06-03
prefer explicit version in Accept header

DerNamenlose:
If the Accept header contains an explicit version and
the Content-Type does not, the explicit version is now
preferred.

This would close issue #746.


commonsensesoftware:
Thank you for taking the time to put together this PR. There a lot of reasons the behavior should **not** change, but I've mostly caved to your thinking for when people do it the _wrong_ way (which hopefully won't be often). The implementation in `6.0` will now return both, if present. They must be **exactly** the same or it will return in a 400. That enables both (if they're the same), neither (e.g. neutral), and one or the other.

Because this change will result in a behavioral change that will not be obvious in a minor version release, I'm going to close out and politely reject this PR. I've linked to the implementation in the discussion to the new code. I believe that will achieve what you want and you'll get it `6.0` - soon. 😉 

Thanks again. It was also a good discussion.

PR #759, opened 2021-07-20
fix: #758

### LINKED ISSUE #758: OData: "The allowed functions are: allFunctions"
As discussed in #757 
image

The reason for the behavior is an invalid enum check in https://github.com/microsoft/aspnet-api-versioning/blob/e16b579d240574053ebd8e7ee38c8686beaee174/src/Common.OData.ApiExplorer/AspNet.OData/Builder/DefaultODataQueryOptionDescriptionProvider.cs#L244-L245

The value passed to the block is `AllowedFunctions.AllFunctions`

spaasis:
A small tweak to fix #758

Could also be tackled by changing to `!context.AllowedFunctions.HasFlag(AllowedFunctions.All)`

Heads up, by the way: some of the version tests fail when the default decimal separator is `,` instead of `.`

### DIFF
--- a/src/Common.OData.ApiExplorer/AspNet.OData/Builder/DefaultODataQueryOptionDescriptionProvider.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/Builder/DefaultODataQueryOptionDescriptionProvider.cs
@@ -242,7 +242,8 @@
             }
 
             if ( context.AllowedFunctions != AllowedFunctions.None &&
-                 context.AllowedFunctions != AllowedFunctions.All )
+                 context.AllowedFunctions != AllowedFunctions.All &&
+                 context.AllowedFunctions != AllowedFunctions.AllFunctions )
             {
 #pragma warning disable CA1308 // Normalize strings to uppercase
                 var functions = context.AllowedFunctions.ToString().ToLowerInvariant();
--- a/src/Common.OData.ApiExplorer/AspNet.OData/Builder/DefaultODataQueryOptionDescriptionProvider.cs
+++ b/src/Common.OData.ApiExplorer/AspNet.OData/Builder/DefaultODataQueryOptionDescriptionProvider.cs
@@ -17,7 +17,7 @@
 #endif
     public class DefaultODataQueryOptionDescriptionProvider : IODataQueryOptionDescriptionProvider
     {
-        const char Space = ' ';
+        private const char Space = ' ';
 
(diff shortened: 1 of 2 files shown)


src/Common.OData.ApiExplorer/AspNet.OData/Builder/DefaultODataQueryOptionDescriptionProvider.cs
```diff
@@ -242,7 +242,8 @@ static void AppendAllowedOptions( StringBuilder description, ODataQueryOptionDes
             }
 
             if ( context.AllowedFunctions != AllowedFunctions.None &&
-                 context.AllowedFunctions != AllowedFunctions.All )
+                 context.AllowedFunctions != AllowedFunctions.All &&
```
commonsensesoftware, line 245: I don't _think_ this line is required at all anymore. I didn't even realize that `all` is a function. I'm pretty sure I mistook `All` for `AllFunctions`. If things still work with only `AllFunctions`, then we should be good.

commonsensesoftware:
For reference, [AllowedFunctions.cs](https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Query/AllowedFunctions.cs) defines these enumerated values; in case something else was missed.

src/Common.OData.ApiExplorer/AspNet.OData/Builder/DefaultODataQueryOptionDescriptionProvider.cs
spaasis: I removed the 'All' check. See the changed test on how it affects the texts.

As it's possible for the user to say `[EnableQuery(AllowedFunctions = AllowedFunctions.All)]`, maybe it would be better to swap to the `HasFlag(AllowedFunctions.All)` approach?

commonsensesoftware:
Thank you for this. This PR and the plethora work that I just merged have diverged quite a bit. I have this fix included so I'm going to close out this PR, but thank you for pointing out this issue and taking the time to provide a fix. The new packages will have this fix too. I'm still waiting on code signing so the change is in, but I still can't publish the `Microsoft.*` packages - yet. Thanks again.

spaasis:
@commonsensesoftware thanks for the update! I checked out the release notes for the v6 - can't wait to try it out! 
Thanks a bunch for your hard work again! 🎂

Issue #763, opened 2021-08-09
Errors while trying to build solution inside a dev container


aleks-ivanov:
I am trying to build the solution inside a dev container with the following config:

- Ubuntu 20.04
- .NET 5
- .NET Core 3.1

Here is the [build log](https://github.com/dotnet/aspnet-api-versioning/files/6955445/build.log.txt).

There are quite a few errors, so am I missing some tooling ?

commonsensesoftware:
You likely have all the required tooling. There isn't anything special you should have to install. Coincidentally, I just ran into the same problem on another project that uses a similar solution/build setup.

The problem is likely two-fold:

- Build scripts should not use literal `/` or `\` 
- Paths are case-sensitive on most Linux-based systems

The last item is likely the primary culprit. I didn't know/recall that Linux file systems are case-sensitive. The standard MSBuild extension points use **Build.Directory.props** and **Build.Directory.targets**. These files exist, but are lowercase. Things work just fine on Windows and most macOS systems. You'll almost certainly run into trouble building on Linux.

Supporting multi-platform builds was never a goal, but I definitely wanted to enable it if possible. This should be an easy fix. I just need to audit the build files. Shouldn't take too long since it's fresh in my mind. I'll consider this a _bug_ since it violates [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). Frankly, I'm surprised this hasn't come up or otherwise been reported before. Maybe it has, but insufficient details were provided. Thanks for reporting it.

aleks-ivanov:
@commonsensesoftware the capitalization of the three Directory.Build files did enable errorless building of all source and test projects.

Here is the [build log](https://github.com/dotnet/aspnet-api-versioning/files/6961782/build_log.txt).

For the running tests however, currently some require mono as a host and this is the [test log](https://github.com/dotnet/aspnet-api-versioning/files/6961787/test_log.txt).

commonsensesoftware:
As noted, this was primarily due to the casing of MSBuild extension files. GitHub Actions are successfully running on Linux for `ms` (formerly `master`) and `main` which are the branches that matter. `main` also builds successfully on Windows through WSL with Ubuntu. End-to-end tests need to happen on Windows due to the requirement of the full .NET Framework. Thanks for helping drive these fixes.

PR #767, opened 2021-08-09
Resolves: Add native GitHub continuous code security and quality analysis

### LINKED ISSUE #766: Add native GitHub continuous code security and quality analysis
- [CodeQL](https://github.com/github/codeql) is an open source code scanning tool for discovering vulnerabilities across a codebase

Pipeline Foundation is **a non-profit initiative** with the sole purpose of giving back to the IT community by assisting OSS projects with DevOps implementations and best practices.

Built with ❤ by [Pipeline Foundation](https://pipeline.foundation).

aleks-ivanov:
- add `codeql-analysis.yml` which automatically enables CodeQL code security and quality scanner. It executes on every push commit, PR, manually and every day at 8:00AM UTC. A scan check can be viewed on commits and PRs. All alerts and fix suggestion can be viewed at **Security** tab -> **Code scanning alerts** -> **CodeQL**

### File changes:

```
```

Resolves #766


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,55 @@
...
+          queries: security-and-quality
+          languages: csharp
+
+      - name: Build solution
+        shell: pwsh
+        run: |
+          $pathToSolution = "ApiVersioning.sln"
+          $buildConfiguration = "Release"
+          $useSharedCompilation = "false"
+          $testProjects = @( Get-ChildItem -Path test/ -Recurse | Where-Object { $_.Extension -match 'shproj' -or $_.Extension -match 'csproj' } | ForEach-Object { $_.FullName } )
```
commonsensesoftware, line 33: This command can be simplified to just:

```powershell
Get-ChildItem -Path test -Include *.csproj -Recurse | Select-Object FullName
```

A couple of supporting points:

- This _assumes_ (or rather _declares_) it needs to run on Windows, but it doesn't _have_ to be restricted to that (in the future)
- Avoid literal `/` or `\` to ensure cross-platform builds; PowerShell will do the _right_ thing
- There is no need to filter _Shared Projects_ (`*.shproj`) separately because they are automatically included in all C# projects (`*.csproj`). _Shared Projects_ do not compile or have their own output.
- There is no need to wrap the results in an array (`@()`) because the results are already an array. There will always be 2+ items.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,55 @@
...
+
+          dotnet nuget locals all --clear
```
commonsensesoftware, line 35: What is the value of clearing the NuGet cache? This doesn't appear to do anything useful, except slow down the build. This _might_ be useful if local, self-referencing packages were used from build output, but this project doesn't do that. Packages are immutable with idempotent behavior.

Slowing down the build isn't the _worse_ thing in the world; however, if someone runs this locally, this is pretty significant side effect that a user/dev may not realize will happen. I'm always leery of anything that makes changes to someone's local system.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,55 @@
...
+
+          # remove one or more test projects, 
+          # so that CodeQL only analyzes the source code
+          dotnet sln $pathToSolution remove $testProjects
```
commonsensesoftware, line 39: This is a bad idea. This makes modifications to the Solution (`*.sln`) file. Yes, it can be undone with source control or thrown away in a temporary build, but that is not intrinsically obvious to someone testing this locally. There are at least two alternatives.

### Alterative 1 : Loop Over Target Projects
I just realized that the query above is filtering out test projects. Inverting the query would allow looping over the target projects without any modification.  For example:

```powershell
Get-ChildItem -Path src -Include *.csproj -Recurse | ForEach-Object { <# use $_.FullName #> }
```

### Alternative 2 : Use Solution Filters
The simpler and _saner_ approach would be to use [Solution Filters](https://docs.microsoft.com/en-us/visualstudio/msbuild/solution-filters) (`*.slnf`). I presume this would be called something like `codeql-analysis.slnf` and only include the projects to be analyzed. All `dotnet` commands would then accept the `*.slnf` file instead of the `*.sln` file.

> NOTE: `*.slnf` files have been natively supported by MSBuild since 16.7 (see [Issue 4097](https://github.com/dotnet/msbuild/issues/4097))

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,55 @@
...
+          dotnet clean $pathToSolution `
+          --configuration $buildConfiguration
+
+          dotnet restore $pathToSolution
+
+          dotnet build $pathToSolution `
+          -property:UseSharedCompilation=$useSharedCompilation `
+          --configuration $buildConfiguration `
+          --no-incremental `
+          --no-restore
```
commonsensesoftware, line 50: Why break the restore process into two different steps? `dotnet build` intrinsically triggers restore unless you specify `--no-restore` as you've done. Removing `--no-restore` and the `dotnet restore` will merge the operations into one. In addition, the `dotnet restore` command does **not** specify the `--configuration` parameter. As unlikely as it is, it is possible that restore behavior could vary between configurations. If it did, these two commands could produce a failure. That is negated by either using `dotnet restore --configuration $buildConfiguration` or merging these two commands into one.

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,55 @@
...
+
+      - name: Perform CodeQL Analysis
+        uses: github/codeql-action/analyze@v1
+
+# Built with ❤ by [Pipeline Foundation](https://pipeline.foundation)
```
commonsensesoftware, line 55: nit: Curious if this is injected or code-generated? I'm not opposed to _advertising_ things that drive you to the _Pit of Success_. 😉

commonsensesoftware (commented):
Overall, looks good. A few basic thoughts and feedback points.

The change to **ActionSelectionContext.cs** is fine. Not sure why code analysis didn't catch this before. I've notice that the results are occasionally miss some items.

.github/workflows/codeql-analysis.yml
aleks-ivanov: This is simply showing our contribution to the project. We'd really appreciate it if you support our non-profit initiative - Pipeline Foundation 🙂

aleks-ivanov:
@commonsensesoftware on top of the other suggested improvements, after testing the suggested changes in the [dev container thread](https://github.com/dotnet/aspnet-api-versioning/issues/763) (the capitalization of `Directory.Build.props` and `Directory.Build.targets`.) and ensuring the source projects are being built without errors on Ubuntu 20.04, we implemented those fixes here and switched the pipeline runner from Windows to Ubuntu for increased speed of execution.

PR #769, opened 2021-08-09
Resolves: Add a CI pipeline with GitHub Actions

### LINKED ISSUE #768: Add a CI pipeline with GitHub Actions
- Create a pipeline for Continuous Integration (build and test validation, test log generation, assemblies generation) with GitHub Actions

Pipeline Foundation is **a non-profit initiative** with the sole purpose of giving back to the IT community by assisting OSS projects with DevOps implementations and best practices.

Built with ❤ by [Pipeline Foundation](https://pipeline.foundation).

aleks-ivanov:
### The pipeline runs:

1. CI
- build and unit tests validation
- debug assemblies and unit tests results generation
- uploading artifacts to pipeline run (checkout the PR run of the pipeline for artifact download)

### File changes:

```
```

### Additions:

We would like to add Continuous Deployment as well, however we know there are internal procedures often incorporated with Microsoft projects for the packaging, so we wanted to confirm first if what we have as an idea is desirable. The CD would include:
- commit-based automated versioning
- NuGet package generation and signing
- deployment to NuGet.org
- creation of GitHub Release

Please let us know if you'd like us to include that, if anything could be done better and/or if any other automation is required by the project! We would be happy to improve it to satisfactory completion 🙂  

Resolves #768


commonsensesoftware:
I've provided a lot more context and feedback in #768. I'm fine to leave this for a bit until the process is flushed out. The PR build process **never** produces artifacts. The release process is **always** manual and doesn't occur often. I stopped doing automatic artifact publishing because I have no one to help verify output and I've made mistakes in the past. Once you publish to NuGet, it's forever. I manually do a sanity-check before publishing. Official releases are usually only a few times a year at this point, so handling it manually is manageable.

Adding a YAML build definition at this point would have no effect and, if properly detected in Azure DevOps, would likely break the existing builds. The proposed build script, however, could be the foundation of what's used after the move to the .NET Foundation.

aleks-ivanov:
> I've provided a lot more context and feedback in #768. I'm fine to leave this for a bit until the process is flushed out. The PR build process **never** produces artifacts. The release process is **always** manual and doesn't occur often. I stopped doing automatic artifact publishing because I have no one to help verify output and I've made mistakes in the past. Once you publish to NuGet, it's forever. I manually do a sanity-check before publishing. Official releases are usually only a few times a year at this point, so handling it manually is manageable.
> 
> Adding a YAML build definition at this point would have no effect and, if properly detected in Azure DevOps, would likely break the existing builds. The proposed build script, however, could be the foundation of what's used after the move to the .NET Foundation.

Any MS signing is perfectly doable and secure within the GitHub ecosystem using the Secrets module and nuget CLI (and soon with .NET 6 even easier and more convenient with the dotnet CLI instead).

And now that the project's build/test processes are verified to work on Linux, the pipeline could be ran on Ubuntu instead of Windows, which would significantly increase the speed of its execution.

Whenever and whatever you decide on the pipeline, we'd be happy to contribute, just give us the heads up 🙂

PR #770, opened 2021-08-10
Resolves: Errors while trying to build solution inside a dev container

### LINKED ISSUE #763: Errors while trying to build solution inside a dev container

aleks-ivanov:
### GitHub Codespaces development container:

- ready-to-start GitHub Codespaces configuration with all necessary tooling

- in addition, it provides basic tools for:
  - .NET development
  - GitHub support
  - overall more pleasant VS Code experience
  
The configuration consists of:

- `"settings":` - a list of VS Code settings to be applied automatically after the Codespace container is created (.editorconfig overrides these)
  - `"workbench.colorTheme": "Default Dark+"` - sets the theme of the VS Code workbench to the `Default Dark+` theme
  - `"terminal.integrated.defaultProfile.linux": "pwsh"` - sets the default VS Code terminal to PowerShell Core

- `extensions:` - a list of VS Code extensions that are automatically installed after the Codespace container is created
  - `"coenraads.bracket-pair-colorizer"` - sets different colors for each nested pair of brackets
  - `"vscode-icons-team.vscode-icons"` - provides a huge set of icons for the VS Code explorer
  - `"editorconfig.editorconfig"` - attempts to override user/workspace settings with those in the .editorconfig
  - `"eamodio.gitlens"` - provides git information directly inside the code
  - `"cschleiden.vscode-github-actions"` and `"redhat.vscode-yaml"` - provide YAML and GitHub Actions support
  - `"bierner.markdown-preview-github-styles"` and `"ban.spellright"` - provide assistance with writing Markdown documentation
  - `"ms-dotnettools.csharp"` and `"VisualStudioExptTeam.vscodeintellicode"` - provide basic Visual Studio tooling
  - `"ms-vscode.powershell"` - provides the functionality of Windows PowerShell ISE inside VS Code
  - `"jmrog.vscode-nuget-package-manager"` - provides use of the NuGet library through the Command Palette
  
- `"postCreateCommand"` - is a string of commands separated by `&&` that execute after the container has been built and the source code has been cloned

- `"build"` - declares the Docker configuration that the container would use to run.

- `Dockerfile`:
  - `"mcr.microsoft.com/vscode/devcontainers/dotnet:0.201.9-5.0"` - the Codespace container runs from an Ubuntu 20.04 image with .NET Core SDK installed (`0.201.9` is the latest .NET Core SDK Docker image tagged version)
  - Additional installed tools:
    - Mono
    - .NET Core SDK 3.1

This GitHub Codespace configuration can also be used locally with the [Remote - Containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) extension for VS Code. It automatically creates and runs a Docker container based on the `devcontainer.json` configuration inside the repo, so anyone could work on the project from any computer, without the need to install anything other than VS Code and Docker.

### File changes:

- Capitalization of `src/directory.build.props`, `src/directory.build.targets`, `test/directory.build.props` to fix the case-sensitivity issues of MSBuild under Linux

```
```

Resolves #763


commonsensesoftware:
I know the changes are small, but this probably should have been 2 PRs. You took the time to fix the file name casing. I have zero issue with that and would take it immediately. I'm not completely opposed to Codespaces support, but I don't remember a discussion about it. As a one many army, I am little finicky about some changes because I ultimately have to support them after it's merged.

I don't really have a problem with having better support for VS Code or Codespaces. I occasionally use VS Code, but not on this project. I don't think I've used anything that couldn't also be done with VS Community Edition, which is free.

The only two significant callouts are:

1. I don't see any need for Windows PowerShell. Any and all support should be PowerShell (formerly Core).
    a. That would virtually guarantee any scripting across platforms will work (though I don't there will ever been any here)
    b. Windows machines can use the `*.msix` installation path so that it doesn't require elevation to install or use
        i. Or it could just be caveat emptor if PowerShell isn't installed
    c. After 6+ years of this project it is **very** unlikely there will be any out-of-band scripting. All build actions **must** be done through the .NET SDK, MSBuild, and/or custom builds (which there are none in this project)
2. Why is Mono specifically needed or is that just a product of the older SDK?
3. How does this handle multi-targeting?
    a. The configuration only targets .NET Core 3.1
    b. The `master` branch will also be support .NET 5.0 (it hasn't been merged yet)
    c. `main` is about to become the new default branch and it will be .NET 6.0+ going forward
    d. How would you test the .NET Framework targets on a non-Windows machine? The full .NET Framework isn't there.
        i. While it's definitely possible, the current test suite setup is not platform-aware. You _should_ be able to compile on a non-Windows machine, but running the test suite for all of the older ASP.NET Web API libraries that target the full .NET Framework will fail on a non-Windows machine methinks

You put in the effort so I want to give create where credit is due.

commonsensesoftware:
This particular change set no longer applies and I'm not even sure if I can use Codespaces within the .NET Foundation organization, but I've incorporated your changes nevertheless. I need a Windows build to run the full .NET Framework unit tests, but I've verified the build on Ubuntu with WSL. Thanks for the tips. You'll see these changes reflected in the new **main** branch.

PR #889, opened 2022-10-06
Fix versioning error when using composite key

mlunelli:
I had an error when try to implement a Get action with a model with composite key, whis is the error:

``` CSHARP
 at Asp.Versioning.ApiExplorer.ODataApiDescriptionProvider.TryMatchModelVersion(ApiDescription description, IReadOnlyList`1 items, IODataRoutingMetadata& metadata)
   at Asp.Versioning.ApiExplorer.ODataApiDescriptionProvider.OnProvidersExecuted(ApiDescriptionProviderContext context)
   at Microsoft.AspNetCore.Mvc.ApiExplorer.ApiDescriptionGroupCollectionProvider.GetCollection(ActionDescriptorCollection actionDescriptors)
   at Microsoft.AspNetCore.Mvc.ApiExplorer.ApiDescriptionGroupCollectionProvider.get_ApiDescriptionGroups()
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.GetSwaggerDocumentWithoutFilters(String documentName, String host, String basePath)
   at Swashbuckle.AspNetCore.SwaggerGen.SwaggerGenerator.<GetSwaggerAsync>d__6.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at Swashbuckle.AspNetCore.Swagger.SwaggerMiddleware.<Invoke>d__4.MoveNext()
   at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
   at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
   at System.Runtime.CompilerServices.TaskAwaiter.GetResult()
```

I added a new example project ( SomeODataOpenApiExampleCompositeKey ) to reproduce the error.

In the second commit you can find a possible fix to the issue.

Thanks
Michele

### DIFF
--- a/asp.sln
+++ b/asp.sln
@@ -197,6 +197,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SomeODataOpenApiExample", "
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SomeOpenApiODataWebApiExample", "examples\AspNet\OData\SomeOpenApiODataWebApiExample\SomeOpenApiODataWebApiExample.csproj", "{AC952FBF-D7DC-4DE4-AD1C-4CEA589034F5}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "SomeODataOpenApiExampleCompositeKey", "examples\AspNetCore\OData\SomeODataOpenApiExampleCompositeKey\SomeODataOpenApiExampleCompositeKey.csproj", "{452D7A02-C198-48AA-8F99-0445653F4792}"
+EndProject
 Global
 	GlobalSection(SolutionConfigurationPlatforms) = preSolution
 		Debug|Any CPU = Debug|Any CPU
@@ -387,6 +389,10 @@ Global
 		{AC952FBF-D7DC-4DE4-AD1C-4CEA589034F5}.Debug|Any CPU.Build.0 = Debug|Any CPU
 		{AC952FBF-D7DC-4DE4-AD1C-4CEA589034F5}.Release|Any CPU.ActiveCfg = Release|Any CPU
 		{AC952FBF-D7DC-4DE4-AD1C-4CEA589034F5}.Release|Any CPU.Build.0 = Release|Any CPU
+		{452D7A02-C198-48AA-8F99-0445653F4792}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{452D7A02-C198-48AA-8F99-0445653F4792}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{452D7A02-C198-48AA-8F99-0445653F4792}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{452D7A02-C198-48AA-8F99-0445653F4792}.Release|Any CPU.Build.0 = Release|Any CPU
 	EndGlobalSection
 	GlobalSection(SolutionProperties) = preSolution
 		HideSolutionNode = FALSE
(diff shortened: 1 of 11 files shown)


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/aspnet-api-versioning?pullRequest=889) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/aspnet-api-versioning?pullRequest=889) before we can accept your contribution.<br/><br/>:x: mlunelli [sign now](https://cla.dotnetfoundation.org/dotnet/aspnet-api-versioning?pullRequest=889)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/aspnet-api-versioning?pullRequest=889) it.</sub>

commonsensesoftware:
As much as I appreciate PRs (because they don't come often), I think there is a misunderstanding of this example project. The **SomeODataOpenApiExample** is meant to highlight how to use OpenAPI with only _some_ OData. Honestly, I struggled to come up with a good name. The reason that composite keys were not working is because not all of the OData features were on or enabled.

There are some people out there that want to use vanilla endpoints, but still have OData query capabilities (without an EDM, etc). This means there is no base class or it's likely `ControllerBase` (as opposed to `ODataController`). That _is_ a supported scenario by the OData libraries, albeit not particularly well-known. There had been a long-time ask to support exploring OData query options for OpenAPI when not using OData. In fact, before it was supported, encountering `ODataQueryOptions` would explode into 1600+ parameters! The **SomeODataOpenApiExample** project is meant to show how that needs to be setup.

If you take all of the bits for the `Book` entity set and add it to the [ODataOpenApiExample](../blob/main/examples/AspNetCore/OData/ODataOpenApiExample), things work as expected. You'll see an entry added for `api/Books/idFirst={keyidFirst},idSecond={keyidSecond}?api-version=1.0` and making a request works as intended.

I'm happy to continue the discussion, but this PR doesn't appear to fix or enhance anything.

mlunelli:
Thanks for your quick reply!

In the [ODataOpenApiExample](https://github.com/dotnet/aspnet-api-versioning/blob/main/examples/AspNetCore/OData/ODataOpenApiExample) if you comment out the line 20 in Program.cs (options.RouteOptions.EnableKeyInParenthesis = false;) you get the error too.

commonsensesoftware:
Hmm... I tested that out and got a different error. The clue is a NRE that happens due to a `null` return value [here](../blob/main/src/AspNetCore/OData/src/Asp.Versioning.OData.ApiExplorer/ApiExplorer/ODataApiDescriptionProvider.cs#L191). This _should_ be impossible.

I think this definitely is a 🐞 , but I don't think it's specific to composite keys. That was simply a symptom. My observations were as follows:

### Works

```c#
options.RouteOptions.EnableKeyInParenthesis = false;
options.RouteOptions.EnableKeyAsSegment = true;
```

### Works

```c#
options.RouteOptions.EnableKeyInParenthesis = true;
options.RouteOptions.EnableKeyAsSegment = false;
```

### Fails

```c#
// this is also the default and same result as if nothing was set
options.RouteOptions.EnableKeyInParenthesis = true;
options.RouteOptions.EnableKeyAsSegment = true;
```

I know that OData is adding different `SelectorModel` instances, when only one is expected. There's even some comments about that in the ASP.NET Core codebase (which simply states that it's never been a problem 😶‍🌫️). This is problematic for API Versioning because the endpoint metadata isn't copied by default. I thought I found all the edge cases and properly copied the metadata over; clearly not.

I think it's probably worth opening a new issue that calls out the bug. I don't have a good description yet. I suspect there are scenarios beyond composites keys that will cause this behavior. The only possible workaround at the moment will be to use either _parenthesis_ or _as segment_ for keys, but not both.

I'll provide some more details once I've had to dig in a little more. Good find. 😃

commonsensesoftware:
This was a great find and discussion. Alas, this PR doesn't address the crux of the problem; it's a bit broader.  I tracked this with a bit more detail in #891. I've put a new PR which will address this issue and be published with `6.2`. Thank you for your help and contribution.

PR #1151, opened 2025-11-26
Support Deprecation header

### LINKED ISSUE #1128: Support for the "Deprecation" Response Header (RFC 9745)
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Is your feature request related to a problem? Please describe the problem.

Supporting the `Deprecation` Response Header and the `Deprecation` Link Relation Type for deprecated API version, as specified in [RFC 9745](https://www.rfc-editor.org/rfc/rfc9745.html)

### Describe the solution you'd like

A configurable _Deprecation Policy_, similar to _Sunset Policy_, like this

```
options.Policies.Deprecation(0.9)
  .Effective(2025, 3, 31)
  .Link("https://developer.example.com/deprecation")
    .Type("text/html");
```

that will produce as response headers

```
Deprecation: @1743379200
Link: <https://developer.example.com/deprecation>; rel="deprecation"; type="text/html"
```

### Additional context

_No response_

MGessinger:
# Add new policy to configure `Deprecation` header field

Add a new `DeprecationPolicy` which emits a `Deprecation` header when a matching API version is used.

## Description

An API may emit a `Deprecation` header to inform clients that a resource is/will become deprecated (but may still be available past that point). The new `DeprecationPolicy` allows users to configure this header, as well as the (optional) associated links. This is analogous to the existing `SunsetPolicy`.

The implementation follows the existing `SunsetPolicy` (and associated classes like `SunsetPolicyBuilder`, `SunsetPolicyManager` etc) as closely as possible. In an attempt to minimize code duplication between the two, common functionality was extracted into shared base classes/interfaces. There is still some duplication left over, but I chose what seemed like a reasonable a middle ground between streamlining the two policies, and keeping the code clear.

At the time of writing this, no tests have been added yet. Since this is a rather large PR with many changed files, I want a second opinion on the implementation before I start pinning everything down in tests. I did at least verify that the solution compiles.

## Design choices
This summarizes the linked issue #1128 

The new policy is configured like this:
```CSharp
options.Policies.Deprecate( 0.9 )
                .Effective( DateTimeOffset.Now.AddDays( 60 ) )
                .Link( "deprecation.html" )
                    .Title( "Deprecation Policy" )
                    .Type( "text/html" );
```
- When a deprecation policy is configured, then it will alwys be omitted, regardless whether the deprecation date is in the future or in the past.
- The RelationType of the configured link must be `deprecation` as per the spec.
- There is no constraint on the provided media type of the link.

Fixes #1128

### DIFF
--- a/src/Abstractions/src/Asp.Versioning.Abstractions/DeprecationPolicy.cs
+++ b/src/Abstractions/src/Asp.Versioning.Abstractions/DeprecationPolicy.cs
@@ -73,6 +73,6 @@ public class DeprecationPolicy
             return true;
         }
 
-        return date < when;
+        return date <= when;
     }
 }
--- a/src/Client/src/Asp.Versioning.Http.Client/System.Net.Http/HttpResponseMessageExtensions.cs
+++ b/src/Client/src/Asp.Versioning.Http.Client/System.Net.Http/HttpResponseMessageExtensions.cs
@@ -19,7 +19,7 @@ public static class HttpResponseMessageExtensions
     private const string Link = nameof( Link );
 
 #if NETSTANDARD1_1
-    private static readonly DateTime UnixEpoch = new DateTime( 1970, 1, 1 );
+    private static readonly DateTime UnixEpoch = new DateTime( 1970, 1, 1, 0, 0, 0, DateTimeKind.Utc );
 #endif
 
     /// <summary>
@@ -82,7 +82,7 @@ public static class HttpResponseMessageExtensions
     public static string ToDeprecationHeaderValue( this DateTimeOffset deprecationDate )
     {
         var unixTimestamp = deprecationDate.ToUnixTimeSeconds();
-        return unixTimestamp.ToString( "'@'0", CultureInfo.CurrentCulture );
+        return unixTimestamp.ToString( "'@'0", CultureInfo.InvariantCulture );
     }
 
     /// <summary>
@@ -100,7 +100,7 @@ public static class HttpResponseMessageExtensions
 
         if ( headers.TryGetValues( Deprecation, out var values ) )
         {
-            var culture = CultureInfo.CurrentCulture;
+            var culture = CultureInfo.InvariantCulture;
             var style = NumberStyles.Integer;
 
             foreach ( var value in values )
--- a/src/AspNet/WebApi/src/Asp.Versioning.WebApi.ApiExplorer/ApiExplorer/VersionedApiExplorer.cs
+++ b/src/AspNet/WebApi/src/Asp.Versioning.WebApi.ApiExplorer/ApiExplorer/VersionedApiExplorer.cs
@@ -238,13 +238,11 @@ public class VersionedApiExplorer : IApiExplorer
         }
 
         var routes = FlattenRoutes( Configuration.Routes ).ToArray();
-        var sunsetPolicyManager = Configuration.GetSunsetPolicyManager();
-        var deprecationPolicyManager = Configuration.GetDeprecationPolicyManager();
 
         foreach ( var apiVersion in FlattenApiVersions( controllerMappings ) )
         {
-            sunsetPolicyManager.TryGetPolicy( apiVersion, out var sunsetPolicy );
-            deprecationPolicyManager.TryGetPolicy( apiVersion, out var deprecationPolicy );
+            SunsetPolicyManager.TryGetPolicy( apiVersion, out var sunsetPolicy );
+            DeprecationPolicyManager.TryGetPolicy( apiVersion, out var deprecationPolicy );
 
             for ( var i = 0; i < routes.Length; i++ )
             {
--- a/src/AspNet/WebApi/test/Asp.Versioning.WebApi.Tests/DefaultApiVersionReporterTest.cs
+++ b/src/AspNet/WebApi/test/Asp.Versioning.WebApi.Tests/DefaultApiVersionReporterTest.cs
@@ -13,8 +13,8 @@ public class DefaultApiVersionReporterTest
     public void report_should_add_expected_headers()
     {
         // arrange
-        var sunsetDate = DateTimeOffset.Now;
-        var deprecationDate = DateTimeOffset.Now;
+        var sunsetDate = DateTimeOffset.UtcNow.AddDays( 2 );
+        var deprecationDate = DateTimeOffset.UtcNow.AddDays( 1 );
         var reporter = new DefaultApiVersionReporter( new TestSunsetPolicyManager( sunsetDate ), new TestDeprecationPolicyManager( deprecationDate ) );
         var configuration = new HttpConfiguration();
         var request = new HttpRequestMessage();
@@ -57,13 +57,11 @@ public class DefaultApiVersionReporterTest
         headers.GetValues( "api-supported-versions" ).Should().Equal( "1.0, 2.0" );
         headers.GetValues( "api-deprecated-versions" ).Should().Equal( "0.9" );
         headers.GetValues( "Sunset" )
-               .Single()
                .Should()
-               .Be( sunsetDate.ToString( "r" ) );
+               .ContainSingle( sunsetDate.ToString( "r" ) );
         headers.GetValues( "Deprecation" )
-               .Single()
                .Should()
-               .Be( $"@{unixTimestamp}" );
+               .ContainSingle( $"@{unixTimestamp}" );
         headers.GetValues( "Link" )
                .Should()
                .BeEquivalentTo( [
--- a/src/AspNetCore/WebApi/test/Asp.Versioning.Http.Tests/DefaultApiVersionReporterTest.cs
+++ b/src/AspNetCore/WebApi/test/Asp.Versioning.Http.Tests/DefaultApiVersionReporterTest.cs
@@ -12,8 +12,8 @@ public class DefaultApiVersionReporterTest
     public void report_should_add_expected_headers()
     {
         // arrange
-        var sunsetDate = DateTimeOffset.Now;
-        var deprecationDate = DateTimeOffset.Now;
+        var sunsetDate = DateTimeOffset.UtcNow.AddDays( 2 );
+        var deprecationDate = DateTimeOffset.UtcNow.AddDays( 1 );
         var reporter = new DefaultApiVersionReporter( new TestSunsetPolicyManager( sunsetDate ), new TestDeprecationPolicyManager( deprecationDate ) );
         var httpContext = new Mock<HttpContext>();
         var features = new Mock<IFeatureCollection>();
@@ -65,17 +65,18 @@ public class DefaultApiVersionReporterTest
 
         headers["api-supported-versions"].Should().Equal( "1.0, 2.0" );
         headers["api-deprecated-versions"].Should().Equal( "0.9" );
-        headers["Sunset"].Single()
-                         .Should()
-                         .Be( sunsetDate.ToString( "r" ) );
-        headers["Deprecation"].Single()
-                              .Should()
-                              .Be( $"@{unixTimestamp}" );
-        headers["Link"].Should()
-                       .BeEquivalentTo( [
-                            "<http://docs.api.com/sunset.html>; rel=\"sunset\"",
-                            "<http://docs.api.com/deprecation.html>; rel=\"deprecation\"",
-                       ] );
+        headers["Sunset"]
+            .Should()
+            .ContainSingle( sunsetDate.ToString( "r" ) );
+        headers["Deprecation"]
+            .Should()
+            .ContainSingle( $"@{unixTimestamp}" );
+        headers["Link"]
+            .Should()
+            .BeEquivalentTo( [
+                "<http://docs.api.com/sunset.html>; rel=\"sunset\"",
+                "<http://docs.api.com/deprecation.html>; rel=\"deprecation\"",
+            ] );
     }
 
     private sealed class TestSunsetPolicyManager : IPolicyManager<SunsetPolicy>
--- a/src/Client/test/Asp.Versioning.Http.Client.Tests/System.Net.Http/HttpResponseMessageExtensionsTest.cs
+++ b/src/Client/test/Asp.Versioning.Http.Client.Tests/System.Net.Http/HttpResponseMessageExtensionsTest.cs
@@ -10,7 +10,7 @@ public class HttpResponseMessageExtensionsTest
     public void read_sunset_policy_should_parse_response()
     {
         // arrange
-        var date = DateTimeOffset.Now;
+        var date = DateTimeOffset.UtcNow;
         var request = new HttpRequestMessage( HttpMethod.Get, "http://tempuri.org" );
         var response = new HttpResponseMessage() { RequestMessage = request };
 
@@ -35,7 +35,7 @@ public class HttpResponseMessageExtensionsTest
     public void read_sunset_policy_should_use_greatest_date()
     {
         // arrange
-        var date = DateTimeOffset.Now;
+        var date = DateTimeOffset.UtcNow;
         var expected = date.AddDays( 14 );
         var request = new HttpRequestMessage( HttpMethod.Get, "http://tempuri.org" );
         var response = new HttpResponseMessage() { RequestMessage = request };
@@ -91,7 +91,7 @@ public class HttpResponseMessageExtensionsTest
     public void read_deprecation_policy_should_parse_response()
     {
         // arrange
-        var date = DateTimeOffset.Now;
+        var date = DateTimeOffset.UtcNow;
         var request = new HttpRequestMessage( HttpMethod.Get, "http://tempuri.org" );
         var response = new HttpResponseMessage() { RequestMessage = request };
 
@@ -116,7 +116,7 @@ public class HttpResponseMessageExtensionsTest
     public void read_deprecation_policy_should_use_smallest_date()
     {
         // arrange
-        var date = DateTimeOffset.Now;
+        var date = DateTimeOffset.UtcNow;
         var expected = date.Subtract( TimeSpan.FromDays( 14 ) );
         var request = new HttpRequestMessage( HttpMethod.Get, "http://tempuri.org" );
         var response = new HttpResponseMessage() { RequestMessage = request };
--- a/src/Client/test/Asp.Versioning.Http.Client.Tests/net#.0/ApiVersionHandlerLoggerTTest.cs
+++ b/src/Client/test/Asp.Versioning.Http.Client.Tests/net#.0/ApiVersionHandlerLoggerTTest.cs
@@ -23,8 +23,8 @@ public class ApiVersionHandlerLoggerTTest
         var date = DateTimeOffset.Now;
         var expected = "API version 1.0 for http://tempuri.org has been deprecated since <unspecified> and will " +
                       $"sunset on {date.ToUniversalTime()}. Additional information: " +
-                      "API Policy (en): http://tempuri.org/policy/en, " +
-                      "API Política (es): http://tempuri.org/policy/es";
+                      "[API Policy (en): http://tempuri.org/policy/en, " +
+                      "API Política (es): http://tempuri.org/policy/es]";
 
         response.Headers.Add( "sunset", date.ToString( "r" ) );
         response.Headers.Add( "link", "<policy/en>; rel=\"sunset\"; type=\"text/html\"; title=\"API Policy\"; hreflang=\"en\"" );
--- a/src/Common/src/Common/DefaultApiVersionReporter.cs
+++ b/src/Common/src/Common/DefaultApiVersionReporter.cs
@@ -104,6 +104,7 @@ public sealed partial class DefaultApiVersionReporter : IReportApiVersions
 
         if ( deprecationPolicyManager.TryResolvePolicy( name, version, out var deprecationPolicy ) )
         {
+            // Only emit a deprecation header if the deprecation policy becomes effective before the sunset date.
             if ( deprecationPolicy.IsEffective( sunsetDate ) )
             {
                 response.WriteDeprecationPolicy( deprecationPolicy );
--- a/src/Common/test/Common.Tests/DeprecationPolicyBuilderTest.cs
+++ b/src/Common/test/Common.Tests/DeprecationPolicyBuilderTest.cs
@@ -51,7 +51,7 @@ public class DeprecationPolicyBuilderTest
     }
 
     [Fact]
-    public void build_should_construct_sunset_policy()
+    public void build_should_construct_deprecation_policy()
     {
         // arrange
         var builder = new DeprecationPolicyBuilder( default, ApiVersion.Default );
@@ -66,6 +66,6 @@ public class DeprecationPolicyBuilderTest
         policy.Should().BeEquivalentTo(
             new DeprecationPolicy(
                 new DateTimeOffset( new DateTime( 2022, 2, 1 ) ),
-                new LinkHeaderValue( new Uri( "http://tempuri.org" ), "sunset" ) ) );
+                new LinkHeaderValue( new Uri( "http://tempuri.org" ), "deprecation" ) ) );
     }
 }


MGessinger:
@dotnet-policy-service agree

commonsensesoftware:
Thanks for this. Just acknowledging that I've seen it. I'll try to put eyes on it ASAP. FYI, I'll be on vacation next week. I'm not ignoring it. 😉

MGessinger:
Have you had any chance to look at this yet?

examples/AspNetCore/OData/ODataOpenApiExample/ConfigureSwaggerOptions.cs
```diff
@@ -50,25 +50,84 @@ private static OpenApiInfo CreateInfoForApiVersion( ApiVersionDescription descri
             text.Append( " This API version has been deprecated." );
         }
 
-        if ( description.SunsetPolicy is { } policy )
+        if ( description.DeprecationPolicy is { } deprecationPolicy )
         {
-            if ( policy.Date is { } when )
+            if ( deprecationPolicy.Date is { } when )
```
commonsensesoftware: Shouldn't these messages expand rather than be replaced? They are complementary with different meanings.

- **Deprecated** - obsolete and not recommended
- **Sunset** - permanently removed

An API _could_ be deprecated forever, but if it does have a sunset policy, then that indicates when it will go away - forever.

src/Abstractions/src/Asp.Versioning.Abstractions/ISunsetPolicyManager.cs
```diff
@@ -5,18 +5,5 @@ namespace Asp.Versioning;
...
-    /// <param name="sunsetPolicy">The applicable <see cref="SunsetPolicy">sunset policy</see>, if any.</param>
-    /// <returns>True if the <paramref name="sunsetPolicy">sunset policy</paramref> was retrieved; otherwise, false.</returns>
-    /// <remarks>If <paramref name="name"/> is <c>null</c>, it is assumed the caller intends to match any sunset
-    /// policy for the specified <paramref name="apiVersion">API version</paramref>. If
-    /// <paramref name="apiVersion">API version</paramref> is <c>null</c>, it is assumed the caller intends to match
-    /// any sunset policy for the specified <paramref name="name"/>.</remarks>
-    bool TryGetPolicy( string? name, ApiVersion? apiVersion, [MaybeNullWhen( false )] out SunsetPolicy sunsetPolicy );
-}
\ No newline at end of file
+public interface ISunsetPolicyManager : IPolicyManager<SunsetPolicy>
```
commonsensesoftware: Is this worth keeping? I'm sure you did this to avoid the breaking change. Since we're at a major version LTS release, I'd be OK with removing `ISunsetPolicyManager`. It's just a backward-compatible, marker interface now. If this were to be introduced as a breaking change, now is probably the time. Historically, few - if any - ever dive down to this level of customization so as long as everything else is source compatible, it's doubtful to be problematic.

I'm also considering doing at least one preview release, which _might_ surface if anyone would be upset by that.

src/Abstractions/src/Asp.Versioning.Abstractions/DeprecationPolicy.cs
```diff
@@ -0,0 +1,77 @@
...
+/// Represents an API version deprecation policy.
+/// </summary>
+public class DeprecationPolicy
+{
+    private readonly LinkList links;
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="DeprecationPolicy"/> class.
+    /// </summary>
+    public DeprecationPolicy()
```
commonsensesoftware: nit: this can be:

```c#
public DeprecationPolicy() => links = new DeprecationLinkList();
```

src/Abstractions/src/Asp.Versioning.Abstractions/DeprecationPolicy.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            links.Add( link );
+        }
+    }
+
+    /// <summary>
+    /// Initializes a new instance of the <see cref="DeprecationPolicy"/> class.
+    /// </summary>
+    /// <param name="link">The link which provides information about the deprecation policy.</param>
+    public DeprecationPolicy( LinkHeaderValue link )
```
commonsensesoftware: nit: this can be:

```c#
public DeprecationPolicy( LinkHeaderValue link ) => links.Add( link );
```

>We're covered by the `null`  check because `Collection<T>` does not allow adding or updating with `null`

src/Abstractions/src/Asp.Versioning.Abstractions/SunsetPolicy.cs
```diff
@@ -2,40 +2,46 @@
...
 public class SunsetPolicy
 {
-    private SunsetLinkList? links;
+    private readonly LinkList links;
 
     /// <summary>
     /// Initializes a new instance of the <see cref="SunsetPolicy"/> class.
     /// </summary>
-    public SunsetPolicy() { }
+    public SunsetPolicy()
```
commonsensesoftware: nit: this can be:

```c#
public SunsetPolicy() => links = new SunsetLinkList();
```

src/Abstractions/src/Asp.Versioning.Abstractions/DeprecationPolicy.cs
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+
+namespace Asp.Versioning;
+
+/// <summary>
+/// Represents an API version deprecation policy.
+/// </summary>
+public class DeprecationPolicy
+{
+    private readonly LinkList links;
```
commonsensesoftware: This should be:

```c#
private readonly DeprecationLinkList links;
```

so that it can be _devirtualized_ and optimized.

I this _should_ also initially be `null` internally so that calls do not pay the cost of allocation if the policy doesn't add any links. This would be a similar pattern used in `SunsetPolicy`. That is also the reason for the `HasLinks` property. `Links.Count == 0` could have just as easily worked externally, but then it always requires an allocation just to find out it's empty.

src/Abstractions/src/Asp.Versioning.Abstractions/LinkList.cs
```diff
@@ -0,0 +1,24 @@
+// Copyright (c) .NET Foundation and contributors. All rights reserved.
+
+namespace Asp.Versioning;
+
+using System.Collections.ObjectModel;
+
+internal abstract class LinkList : Collection<LinkHeaderValue>
```
commonsensesoftware: ❓ The only branching logic is the _relation type_ value. Would it be better and simpler to just pass the relation type down rather than subclassing this? That would enable a single, `sealed class`. For example:

```c#
internal static class Format
{
    // The relation type for a {0} policy link must be "{0}".
    internal static readonly CompositeFormat InvalidRelationType = CompositeFormat.Parse( SR.InvalidRelationType );
}

internal sealed class LinkList( string relationType ) : Collection<LinkHeaderValue>
{
	private readonly string relationType = relationType;

	protected override void InsertItem( int index, LinkHeaderValue item )
    {
        EnsureRelationType( item );
        base.InsertItem( index, item );
    }

    protected override void SetItem( int index, LinkHeaderValue item )
    {
        EnsureRelationType( item );
        base.SetItem( index, item );
    }

    private void EnsureRelationType( LinkHeaderValue item )
    {
        if ( !item.RelationType.Equals( relationType, StringComparison.OrdinalIgnoreCase ) )
        {
        	var message = string.Format( CultureInfo.CurrentCulture, Format.InvalidRelationType, relationType );
            throw new ArgumentException( message, nameof( item ) );
        }
    }
}
```

Finally, the fields would be implemented with:

```c#
private readonly LinkList links = new( "sunset" );
```

or

```c#
private readonly LinkList links = new( "deprecation" );
```

respectively.

src/AspNet/WebApi/src/Asp.Versioning.WebApi/System.Net.Http/HttpResponseMessageExtensions.cs
```diff
@@ -35,6 +38,31 @@ public static void WriteSunsetPolicy( this HttpResponseMessage response, SunsetP
...
+    /// <param name="response">The <see cref="HttpResponseMessage">HTTP response</see> to write to.</param>
+    /// <param name="deprecationPolicy">The <see cref="DeprecationPolicy">deprecation policy</see> to write.</param>
+    public static void WriteDeprecationPolicy( this HttpResponseMessage response, DeprecationPolicy deprecationPolicy )
+    {
+        ArgumentNullException.ThrowIfNull( response );
+        ArgumentNullException.ThrowIfNull( deprecationPolicy );
+
+        var headers = response.Headers;
+
+        if ( deprecationPolicy.Date.HasValue )
```
commonsensesoftware: This can be pattern matched with:

```c#
if ( deprecationPolicy.Date is { } when )
```

FYI, this **cannot** be matched as:

```c#
if ( deprecationPolicy.Date is var when )
```

because `when` would be coerce to `DateTime?` whereas `{ }` does not allow `null`.

src/AspNet/WebApi/src/Asp.Versioning.WebApi/System.Net.Http/HttpResponseMessageExtensions.cs
```diff
@@ -35,6 +38,31 @@ public static void WriteSunsetPolicy( this HttpResponseMessage response, SunsetP
...
+        {
+            long unixTimestamp;
+            DateTimeOffset deprecationDate = deprecationPolicy.Date.Value;
+
+            unixTimestamp = (int) deprecationDate.Subtract( unixEpoch ).TotalSeconds;
+
+            headers.Add( Deprecation, $"@{unixTimestamp}" );
```
commonsensesoftware: This block can be simplified to:

```c#
var timestamp = when.Subtract( DateTime.UnixEpoch ).TotalSeconds;
headers.Add( Deprecation, timestamp.ToString( "'@'0") );
```

src/Client/src/Asp.Versioning.Http.Client/System.Net.Http/HttpResponseMessageExtensions.cs
```diff
@@ -69,6 +70,67 @@ public static SunsetPolicy ReadSunsetPolicy( this HttpResponseMessage response )
...
+            var culture = CultureInfo.CurrentCulture;
+
+            foreach ( var value in values )
+            {
+                var split = value.Trim( '@' );
+                if ( long.TryParse( split, out var unixTimestamp ) )
+                {
+                    DateTimeOffset parsed;
+#if NETSTANDARD
+                    parsed = new DateTime(1970, 1, 1) + TimeSpan.FromSeconds(unixTimestamp);
```
commonsensesoftware: `DateTimeOffset.FromUnixTimeSeconds` is available in .NET Standard 1.3+ so this condition can be revised to `#if NETSTANDARD1_1` so that .NET Standard 2.0 takes the better path.

src/Client/src/Asp.Versioning.Http.Client/System.Net.Http/HttpResponseMessageExtensions.cs
```diff
@@ -69,6 +70,67 @@ public static SunsetPolicy ReadSunsetPolicy( this HttpResponseMessage response )
...
+        var date = default( DateTimeOffset );
+        DeprecationPolicy policy;
+
+        if ( headers.TryGetValues( Deprecation, out var values ) )
+        {
+            var culture = CultureInfo.CurrentCulture;
+
+            foreach ( var value in values )
+            {
+                var split = value.Trim( '@' );
```
commonsensesoftware: We don't want to use `String.Trim` here because it will allocate. We can do the test by just checking the first character. Also, as strange as it might sound, a header _can_ have an empty value.

The following should work for all platforms without allocating:

```c#
if ( value.Length < 2 || value[0] != '@' )
{
    continue;
}
```

src/Client/src/Asp.Versioning.Http.Client/System.Net.Http/HttpResponseMessageExtensions.cs
```diff
@@ -69,6 +70,67 @@ public static SunsetPolicy ReadSunsetPolicy( this HttpResponseMessage response )
...
+                if ( long.TryParse( split, out var unixTimestamp ) )
```
commonsensesoftware: Unfortunately, I think this is another place where a compiler directive is required because the newer versions of .NET can use a span here.

```c#
#if NETSTANDARD
if ( long.TryParse( value.Substring( 1 ), out var unixTimestamp ) )
#else
if ( long.TryParse( value.AsSpan()[1..], out var unixTimestamp ) )
#endif
```

src/Common/src/Common/ApiVersioningPolicyBuilder.cs
```diff
@@ -10,13 +10,18 @@ namespace Asp.Versioning;
...
     public virtual IReadOnlyList<T> OfType<T>() where T : notnull
     {
         if ( typeof( T ) == typeof( ISunsetPolicyBuilder ) && sunsetPolicies != null )
         {
-            return ( sunsetPolicies.Values.ToArray() as IReadOnlyList<T> )!;
+            return sunsetPolicies.Values.Cast<T>().ToArray();
+        }
+        else if ( typeof( T ) == typeof( IDeprecationPolicyBuilder ) && deprecationPolicies != null )
+        {
+            return deprecationPolicies.Values.Cast<T>().ToArray();
```
commonsensesoftware: Same here. I expect `.ToArray()` to be more succinct and perform better

src/Common/src/Common/DeprecationPolicyBuilder.cs
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+    private sealed class DeprecationLinkBuilder : LinkBuilder, ILinkBuilder
+    {
+        protected override string RelationType => "deprecation";
+
+        private readonly DeprecationPolicyBuilder policyBuilder;
+
+        public DeprecationLinkBuilder( DeprecationPolicyBuilder policy, Uri linkTarget )
+            : base( linkTarget )
```
commonsensesoftware: nit: this could be `: base( linkTarget ) = policyBuilder = policy;`

src/Common/src/Common/SunsetPolicyManager.cs
```diff
@@ -5,46 +5,5 @@ namespace Asp.Versioning;
...
-            var key = new PolicyKey( builder.Name, builder.ApiVersion );
-
-            mapping[key] = policy;
-        }
-
-        return mapping;
-    }
-}
\ No newline at end of file
+public partial class SunsetPolicyManager : PolicyManager<SunsetPolicy, SunsetPolicyBuilder>, ISunsetPolicyManager
```
commonsensesoftware: ❓ If we ditch `ISunsetPolicyManager`, this isn't needed anymore, right?

src/Abstractions/src/Asp.Versioning.Abstractions/ISunsetPolicyBuilder.cs
```diff
@@ -33,17 +21,4 @@ public interface ISunsetPolicyBuilder
     /// sunset policy is applied.</param>
     /// <returns>The current <see cref="ISunsetPolicyBuilder">sunset policy builder</see>.</returns>
     ISunsetPolicyBuilder Effective( DateTimeOffset sunsetDate );
```
commonsensesoftware: If this were to return `IPolicyBuilder<TPolicy>` with slightly more generic wording for the date parameter and comments, `Effective` could be rolled up into the `IPolicyBuilder<TPolicy>` interface and get rid of this interface, no?

src/Client/src/Asp.Versioning.Http.Client/net#.0/ILoggerExtensions.cs
```diff
@@ -70,16 +70,10 @@ static partial void NewApiVersionAvailable(
     private static string FormatDate( DateTimeOffset? date ) =>
         date.HasValue ? date.Value.ToString( CultureInfo.CurrentCulture ) : "<unspecified>";
 
-    private static string[] FormatLinks( SunsetPolicy sunsetPolicy )
+    private static string[] FormatLinks( IList<LinkHeaderValue> links )
     {
-        if ( !sunsetPolicy.HasLinks )
```
commonsensesoftware: This is exactly the use to for `HasLinks` to test if there are any links, but without allocating. Is this expected to be handled at a higher level now?

src/Abstractions/src/Asp.Versioning.Abstractions/SunsetPolicy.cs
```diff
@@ -2,40 +2,46 @@
...
 
-using System.Collections.ObjectModel;
-
 /// <summary>
 /// Represents an API version sunset policy.
 /// </summary>
 public class SunsetPolicy
 {
-    private SunsetLinkList? links;
+    private readonly LinkList links;
```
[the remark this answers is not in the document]
MGessinger: Initially, my motivation was to localize the initialization of `links` to just one place (especially in light of the now new constructor which takes a string argument). But I understand the optimization - not something I've thought about before, but it makes sense.
I chose a different approach now, which I think combines the benefits of on-demand allocation and a single source of initialization.

src/Abstractions/src/Asp.Versioning.Abstractions/SunsetPolicy.cs
```diff
@@ -56,23 +62,11 @@ public SunsetPolicy() { }
     /// <remarks>If a link is provided, generally only one link is necessary; however, additional
     /// links might be provided for different languages or different formats such as a HTML page
     /// or a JSON file.</remarks>
-    public IList<LinkHeaderValue> Links => links ??= new();
+    public IList<LinkHeaderValue> Links => links;
 
-    private sealed class SunsetLinkList : Collection<LinkHeaderValue>
+    internal sealed class SunsetLinkList : LinkList
```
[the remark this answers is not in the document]
MGessinger: Initially, it needed to be internal because the base class LinkList is internal. But now this whole subclass is gone.

src/Abstractions/src/Asp.Versioning.Abstractions/ISunsetPolicyBuilder.cs
MGessinger: Currently, I can write something like `policyBuilder.Effective( DateTimeOffset.Now ).Link(/*...*/)`. If I change the return type of `Effective` to be the base interface `IPolicyBuilder`, then this chained call only works if I pull `Link` into the base interface as well. To do that, I need to pull the implementation of `Link` from `SunsetPolicyBuilder` into the base `PolicyBuilder`, and to do that I need to get rid of `SunsetLinkBuilder`.
That seems like an awfully big refactoring to get rid of an interface, but maybe I'm missing an easy way out.

MGessinger: One thiing I could do is to pull `Effective` and `Link` into `IPolicyBuilder` but leave the implementation of `Link` in `PolicyBuilder<TPolicy>` as `abstract`. That would avoid most of the refactoring needed to make `Link` work while winning the benefit related to `Effective`.
But is it right to have `Link` in the base `PolicyBuilder`? For the policies that exist currently, yes. But will there be other policies?

src/Client/src/Asp.Versioning.Http.Client/net#.0/ILoggerExtensions.cs
MGessinger: My goal here was to prepare FormatLinks for use with other policies. But I understand the logic, so I added another overload to hopefully get the best of both worlds (early return if there are no links, and a method that is entirely independent of the SunsetPolicy type)

MGessinger:
Thanks for the thorough review! I went through everything and implemented all your suggestions. Hopefully this is in line with what you had in mind. Two or three comments are still open because it wasn't totally clear to me what the best course of action is.

commonsensesoftware:
Looking good. I'll be traveling this weekend. With the holiday and traveling, it might be next week before I get to take a serious look. I just wanted you to know that am looking and your efforts are appreciated. 😉

MGessinger:
Bump. Are you still traveling?

From your previous comment, it seemed that this was at least going in the right direction, so in the meantime I'm going to get started on getting the tests green again and adding any tests that may be missing.

commonsensesoftware:
I am traveling again, but I'm working remote. I'll get eyes on this tonight or tomorrow. Thanks for poking the bear. ;)

src/Abstractions/src/Asp.Versioning.Abstractions/IDeprecationPolicyBuilderExtensions.cs
commonsensesoftware: Correct me if I'm wrong, but this looks/feels like we lost the extensions for `ISunsetPolicyManager`, no? Sunset policies can still build links and effective dates, even if they have a slightly different meaning from deprecation.

Based on the other thread on `Link` and `Effective`, I'm wondering if we're missing a couple of interfaces so we can describe the behavior independently and compose them. This would retain the flexibility of separate interface evolution, keep the 2 distinct interfaces, but still allow for a unified set of extension methods.

I'm not settled on the names, but I could imagine this being refactored into something like:

```c#
public interface ILinkSource
{
	ILinkBuilder Link( Uri linkTarget );
}

public interface IPolicyEffectiveDate
{
	void Effective( DateTimeOffset date );
}

public interface IDeprecationPolicyBuilder : IPolicyBuilder<DeprecationPolicy>, ILinkSource, IPolicyEffectiveDate { }
public interface ISunsetPolicyBuilder : IPolicyBuilder<SunsetPolicy>, ILinkSource, IPolicyEffectiveDate { }

public static class IPolicyBuilderExtensions
{
    public static ILinkBuilder Link( this ILinkSource builder, string linkTarget )
    {
        ArgumentNullException.ThrowIfNull( builder );
        return builder.Link( new Uri( linkTarget, UriKind.RelativeOrAbsolute ) );
    }

    public static TBuilder Effective<TBuilder>( this TBuilder builder, int year, int month, int day )
        where TBuilder : notnull, IPolicyEffectiveDate
    {
        ArgumentNullException.ThrowIfNull( builder );
        builder.Effective( new DateTimeOffset( new DateTime( year, month, day ) ) );
        return builder;
    }
}
```

Thoughts?

src/Client/src/Asp.Versioning.Http.Client/net#.0/ILoggerExtensions.cs
commonsensesoftware: Shouldn't these log messages now include the date the API was deprecated if we know it from the deprecation policy?

commonsensesoftware:
@MGessinger One other important design aspect I wanted to think about and discuss is what we think the intersection of the new _deprecation policy_ should be the existing APIs and metadata. As an example, the `IApiVersionProvider` and `ApiVersionProviderOptions` allow specifying an API version is deprecated.

A user can specify:

```c#
[ApiController]
[ApiVersion(0.9, Deprecated = true)]
[Route("[controller]")]
public class ExampleController
{
    [HttpGet]
    public string Get() => Ok("Demo");
}
```

This controller expresses that it supports version `0.9`, but that version is now deprecated. What it does not say is _when_ it was deprecated.

Conversely,

```c#
options.Policies.Deprecate( 0.9 )
                .Effective( DateTimeOffset.Now.AddDays( 60 ) )
```

indicates that all `0.9` APIs will be deprecated 60 days from now. The mere configuration of this date implies that all API versions will be deprecated after a specific date. This would hold true even if APIs do not specify `Deprecated = true`.

The question becomes _"How do we think about precedence?"_.

I _think_ the precedence would be ranked as:

1. A deprecation policy has been specified for all APIs of a version
    a. This has the advantage of never requiring developers to apply `Deprecated = true` to endpoints
    b. It _might_ make it more difficult in reporting
2. A deprecation policy with a specific name
3. A deprecation policy with a specific version and name combination
4. An endpoint decorated with a deprecated API version
5. All API endpoints of the same endpoint have been annotated as deprecated

I suspect there's some checks in the implementation that have been missed. Today, things are solely based on `#4` and `#5`. Those cases would now have to take into account the other conditions. The goal should be to make sure the two methods of expressing deprecation are as congruent as possible with a low chance of making a mistake.

MGessinger:
I think that precedence makes sense. Although I don't think it matters too much, since the different ways of declaring deprecation aren't particularly conflicting: You can declare deprecation one of various ways, but as soon as one of them is active, your endpoint is deprecated. You can't configure it deprecated one way, and simultaneously configure it "not deprecated" another way.
Nonetheless, the order you proposed seems right and I'll get that implemented. Could you point me in the right direction, where I can find this code?

commonsensesoftware:
The place you're looking for is effectively here:

https://github.com/dotnet/aspnet-api-versioning/blob/a679121ac275ad207ddb65ecc5a56d5e1e963b7a/src/AspNetCore/WebApi/src/Asp.Versioning.Mvc.ApiExplorer/Internal/DescriptionProvider.cs#L72

After all deprecated versions are collated and they aren't supported anywhere, they become deprecated in their entirety. This flows down this path:

https://github.com/dotnet/aspnet-api-versioning/blob/a679121ac275ad207ddb65ecc5a56d5e1e963b7a/src/AspNetCore/WebApi/src/Asp.Versioning.Mvc.ApiExplorer/Internal/DescriptionProvider.cs#L22

Now that I'm explaining it thorough, I'm thinking this is a non-issue. The deprecation policy doesn't say you're deprecated; it just says the policy. This is the same as the sunset policy. The policy doesn't mean you're sunset, it's just the policy. Deprecation can be in the past (e.g. since) or future (e.g. will be). In fact, it doesn't even matter if the API deprecated to apply the policy.

The more I write, the more I think I may have over-thought it.

commonsensesoftware:
I think this is all solid. I have the new OpenAPI stuff done; it's a lot. I didn't want to make it painful for you so I've been working in a parallel branch. I'd like to get this merged so I can rebase. I'd like to get an initial preview release this week. I'm happy to continue the discussion or do some more tweaking and polishing before the final release.

I'm about ready to push the button now, but I see some tests appear to be failing. Is that expected? I'll hold off until you confirm, but otherwise I think we're at a good checkpoint.

MGessinger:
Failing tests are definitely not expected. I'll look into those as well as whatever the bot found some time later today and then I'll give you the go once I'm happy!

MGessinger:
Tests should be fixed now!

> The deprecation policy doesn't say you're deprecated; it just says the policy

I didn't totally follow on this point, but I'm willing to take your word for it. You mentioned publishing a preview release anyway, I expect that it might surface the most obvious oversights. And if some other constellation shows up later, we can fix it then.

For the time being, assuming the tests run through successfully this time, I think this can be merged.

PR #1166, opened 2026-02-27
Fix Minimal API endpoints not being part of OpenAPI document in ASP.Versioning v10 preview

### LINKED ISSUE #1165: Minimal API endpoints produce empty paths and schemas in OpenAPI documents
## Description

When using minimal APIs with `AddApiVersioning().AddApiExplorer().AddOpenApi()`, the generated OpenAPI documents (e.g. `/openapi/v1.json`) have an **empty `paths` object and no schemas**, even though the document itself is otherwise valid (correct `info.version`, versioning metadata, etc.).

The same configuration with **controllers** works correctly — OpenAPI documents are fully populated with paths and schemas.

## Steps to Reproduce

1. Run the `MinimalOpenApiExample` sample project
2. Navigate to `/openapi/v1.json`
3. Observe empty `paths: {}` and no schemas

**Expected**: paths and schemas for Orders and People endpoints  
**Actual**: empty paths, no schemas

For comparison, run the `OpenApiExample` (controllers) project — its `/openapi/v1.json` has fully populated paths and schemas.

## Root Cause

`AddApiExplorerServices()` in `Asp.Versioning.Mvc.ApiExplorer` registers `AddMvcCore().AddApiExplorer()`, which only registers `DefaultApiDescriptionProvider` — the provider that discovers **controller actions** via `IActionDescriptorCollectionProvider`.

For minimal APIs, the built-in ASP.NET Core `EndpointMetadataApiDescriptionProvider` (registered by `AddEndpointsApiExplorer()`) is responsible for discovering minimal API endpoints via `EndpointDataSource`. This provider is **never registered** by the API versioning library.

The downstream `VersionedApiDescriptionProvider` already supports minimal API descriptions (it has `TryUpdateControllerRouteValueForMinimalApi()` and correctly handles non-`ControllerActionDescriptor` actions), but it never receives any minimal API descriptions because they're never generated upstream.

The `OpenApiDocumentService` (Microsoft's internal class) consumes `IApiDescriptionGroupCollectionProvider` which aggregates all `IApiDescriptionProvider` results. Since `EndpointMetadataApiDescriptionProvider` is missing, minimal API endpoints are invisible.

### Proposed Fix

Add `services.AddEndpointsApiExplorer()` in `AddApiExplorerServices()` alongside the existing `services.AddMvcCore().AddApiExplorer()`:

```csharp
private static void AddApiExplorerServices( IApiVersioningBuilder builder )
{
    builder.AddMvc();

    var services = builder.Services;

    services.AddMvcCore().AddApiExplorer();
    services.AddEndpointsApiExplorer(); // <-- register minimal API endpoint discovery
    // ...
}
```

- `AddEndpointsApiExplorer()` is in the shared framework (`Microsoft.AspNetCore.Routing`) — no new package reference needed
- Uses `TryAddEnumerable` internally — safe to call multiple times
- `EndpointMetadataApiDescriptionProvider` runs at `Order = -1100`, before `VersionedApiDescriptionProvider` (`Order = 0`) — correct sequence
- For controller-only apps, it's a no-op (produces no descriptions for controller endpoints)

## Related: Test `minimal_api_should_generate_expected_open_api_document`

The existing test `minimal_api_should_generate_expected_open_api_document` in `AcceptanceTest.cs` appears to validate minimal API OpenAPI generation, but it actually registers **both** controllers (`AddControllers().AddApplicationPart(...)`) **and** minimal API endpoints. The expected fixture `v1.json` only contains the controller path `/Test` — the minimal API path `/test/{id}` is silently absent. The test passes because it validates against controller output, masking this bug.

This test should be restructured to:
1. Remove controller registration so it tests only minimal API endpoint discovery
2. Use a dedicated fixture with the expected minimal API paths
3. A separate mixed test could verify both controllers and minimal APIs appear together

sander1095:
## Summary

Fixes #1165

`AddApiExplorerServices()` only called `AddMvcCore().AddApiExplorer()`, which registers `DefaultApiDescriptionProvider` — a provider that only discovers **controller-based** endpoints. When using minimal APIs, `OpenApiDocumentService` queries `IApiDescriptionGroupCollectionProvider` which aggregates from all registered `IApiDescriptionProvider` instances. Without `EndpointMetadataApiDescriptionProvider`, minimal API endpoints were never discovered, resulting in empty `paths` and no `schemas` in the generated OpenAPI documents.

## Fix

Added `services.AddEndpointsApiExplorer()` in `AddApiExplorerServices()`, which registers `EndpointMetadataApiDescriptionProvider`. This call is safe to invoke multiple times as it uses `TryAddEnumerable` internally.

## Changes

- **`IApiVersioningBuilderExtensions.cs`**: Added `services.AddEndpointsApiExplorer()` after `services.AddMvcCore().AddApiExplorer()` (1 line)
- **`AcceptanceTest.cs`**: 
  - Rewrote `minimal_api_should_generate_expected_open_api_document` to test a pure minimal-API-only scenario (no controllers). Application parts are cleared to prevent MVC auto-discovery of test controllers. Asserts that `/test/{id}` path is present with correct `id` and `api-version` parameters.
  - Added `mixed_api_should_generate_expected_open_api_document` to verify both controller (`/Test`) and minimal API (`/minimal/{id}`) endpoints appear together in the same OpenAPI document.


sander1095:
I am unable to view the error logs of the build due to a required login.

commonsensesoftware:
CI builds should be fully anonymous. It looks like 2 failed tests.

https://dev.azure.com/aspnet-api-versioning/build/_build/results?buildId=619&view=logs&j=12f1170f-54f2-53f3-20dd-22fc7dff55f9&t=9c939e41-62c2-5605-5e05-fc3554afc9f5

I'm currently traveling.  I'll return home Tuesday.

sander1095:
The tests fail due to 2 culture issues:

- I run on Windows, and the newlines in the expected vs actual OpenAPI generated document differ.
- The dates expect an (american) 1/1/2026 notation. This can be fixed by using the `AssumeCulture` attribute

sander1095:
The tests have been fixed! 🥳

commonsensesoftware:
>I run on Windows, and the newlines in the expected vs actual OpenAPI generated document differ.

Strange that would be an issue. I too run on Windows and so does the build agent. 🤷🏽 

>The dates expect an (american) 1/1/2026 notation...

There are other kinds of formats and notations? 🤔 🤣 

I've been hit by this kind of thing before. This is a reasonable fix. I try to stay out of the localization biz, but this was a valid use case to have some default text. Almost crazy to think there are different, _cultural_ ways of the writing the same text in the same language. 😄

